PR #4, opened 2017-04-24
WIP Serialize params

gavin-norman-sociomantic:


src/swarm/neo/client/mixins/ClientCore.d
```diff
@@ -581,10 +581,10 @@ template ClientCore ( )
 
     private RequestId assign ( R, P ) ( P params )
     {
-        static assert(is(P == R.UserSpecifiedParams));
+        static assert(is(P == Const!(R.UserSpecifiedParams)));
```
mihails-strasuns-sociomantic: I'd use `:` instead of `==` here, as both mutable and immutable are implicitly convertible to const.

src/swarm/neo/client/mixins/RequestCore.d
```diff
@@ -178,7 +178,12 @@ public template RequestCore ( RequestType request_type_, ubyte request_code,
     {
         public Args args;
 
-        public SerializableReferenceType!(Notifier) notifier;
+        ubyte[Notifier.sizeof] notifier;
```
mihails-strasuns-sociomantic: Not immediately obvious, why was this needed?

src/swarm/neo/client/mixins/RequestCore.d
```diff
@@ -190,13 +195,28 @@ public template RequestCore ( RequestType request_type_, ubyte request_code,
 
     public static struct Context
     {
+        import ocean.util.serialize.contiguous.Serializer;
+        import ocean.util.serialize.contiguous.Deserializer;
+
+        private void[] user_specified_params_blob;
```
mihails-strasuns-sociomantic: the host struct is reused, right?

mihails-strasuns-sociomantic: maybe define it as `Contiguous!(UserSpecifiedParams)` if yes:

```D
        import ocean.util.serialize.contiguous.Util;
        
        public Contiguous!(UserSpecifiedParams) user_specified_params;

        public void serializeUserSpecifiedParams (
            Const!(UserSpecifiedParams) params )
        {
            copy(params, this.user_specified_params);
        }
```

mihails-strasuns-sociomantic (commented):
Resulting code change looks pretty clean!

src/swarm/neo/client/mixins/RequestCore.d
gavin-norman-sociomantic: Before, setting up a struct instance looked like:
```D
Request.UserSpecifiedParams params;
...
params.notifier.set(user_notifier); // mutates params
```
If `params` is declared const, it's not possible to call `params.notifier.set(...);`.

src/swarm/neo/client/mixins/RequestCore.d
david-eckardt-sociomantic: We thought of that. However, `Contiguous.data` is not a public field so the de/serialiser would mess with private property. Do you think that would be better?

gavin-norman-sociomantic: >the host struct is reused, right?

A `Context` instance is constructed and serialized. Each request handler is passed the serialized buffer and deserializes a `Context` struct again.

mihails-strasuns-sociomantic: > However, Contiguous.data is not a public field so the de/serialiser would mess with private property

`copy` does it for you - https://github.com/sociomantic-tsunami/ocean/blob/v2.x.x/src/ocean/util/serialize/contiguous/Util.d#L68 (though `const` fix for `copy` was merged very recently and may not be available for swarm :X)

mihails-strasuns-sociomantic: @gavin-norman-sociomantic so that means one GC allocation per each request or?

david-eckardt-sociomantic: Sorry, I misunderstood.

gavin-norman-sociomantic: >so that means one GC allocation per each request or?

Ah shit, you're right. That's no good.

gavin-norman-sociomantic:
Updated to use shared/reused serialization buffer.

src/swarm/neo/client/mixins/RequestCore.d
```diff
@@ -235,10 +304,10 @@ public template RequestCore ( RequestType request_type_, ubyte request_code,
 
     ***************************************************************************/
 
-    private static void notify ( ref UserSpecifiedParams params,
+    private static void notify ( UserSpecifiedParams* params,
```
david-eckardt-sociomantic, line 307: Why this change?

gavin-norman-sociomantic: Just for convenience. `user_params` returns a pointer to a struct now, so usage is simpler if `notify` accepts the same type.

nemanja-boric-sociomantic: Maybe in the next major we could make this `Const!(UserSpecifiedParams)*` params? So that would make a breaking change, since the protos must update their `Notifier` definitions, to accept `Args` by const, but users would just have the read-only view of their args in the notifiers.

gavin-norman-sociomantic: Yes, I thought of that too.

gavin-norman-sociomantic:
Closing. To be replaced with a proper version.

PR #9, opened 2017-04-28
const-correct output methods

david-eckardt-sociomantic:
Change the types of output buffers to `const`, all the way from the `RequestOnConn` API methods down to the socket output function.

### DIFF
--- a/src/swarm/neo/protocol/socket/uio_const.d
+++ b/src/swarm/neo/protocol/socket/uio_const.d
@@ -1,0 +1,165 @@
+/*******************************************************************************
+
+    D-`const`-correct vector/gather socket output with optional `SIGPIPE`
+    suppression.
+
+    Copyright: Copyright (c) 2015-2017 sociomantic labs GmbH. All rights reserved
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+
+*******************************************************************************/
+
+module swarm.neo.protocol.socket.uio_const;
+
+import core.sys.posix.sys.types: ssize_t;
+
+import core.sys.posix.sys.socket;
+
+import ocean.transition;
+
+/*******************************************************************************
+
+    D-`const`-correct output version of `iovec`. It is used like `iovec` for
+    functions that only read the data referenced by `iov_base`, i.e. output
+    functions. It fixes the following problem:
+
+    ```
+        int socket_fd;
+
+        // Define some read-only data to pass to sendv.
+        auto a = "Hello", b = "World";
+
+        // Declare an output vector
+        iovec[2] output_vec;
+
+        // Set the output vector components to reference the read-only data.
+        output_vec[0].iov_base = a.ptr;     // Problem
+        output_vec[0].iov_len  = a.length;
+        output_vec[1].iov_base = b.ptr;     // Problem
+        output_vec[1].iov_len  = b.length;
+
+        // Send the output data i.e. a ~ b through the socket.
+        auto bytes_sent = sendv(socket_fd, output_vec);
+    ```
+
+    The problem here is that the type of `iovec.iov_base` is (mutable) `void*`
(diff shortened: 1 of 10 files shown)


mihails-strasuns-sociomantic (approved):
All `cast(Const!(void*))` bits looks somewhat scary but I couldn't find anything practically wrong.

src/swarm/neo/protocol/socket/uio_const.d
```diff
@@ -0,0 +1,78 @@
...
+{
+    Const!(msghdr) m = {
+            msg_iov:    cast(Const!(iovec*))data.ptr,
+            msg_iovlen: data.length
+    };
+    return sendmsg(socket_fd, &m, flags);
+}
+
+// Ensure binary compatibility of `iovec_const` with `iovec`.
+unittest
```
gavin-norman-sociomantic, line 144: Would be cool to have a template in ocean: `BinaryLayoutIdentical`, or some such. (Future suggestion.)

src/swarm/neo/protocol/socket/uio_const.d
```diff
@@ -0,0 +1,78 @@
...
+
+module swarm.neo.protocol.socket.uio_const;
+
+import core.sys.posix.sys.types: ssize_t;
+
+import core.sys.posix.sys.socket;
+
+import ocean.transition;
+
+/// D-`const`-correct output version of `iovec`.
```
gavin-norman-sociomantic: Would be nice to document why this is declared and how it is used.

src/swarm/neo/protocol/socket/uio_const.d
```diff
@@ -0,0 +1,78 @@
...
+
+    Returns:
+        0 on success or -1 on failure, in which case `errno` is set accordingly.
+
+*******************************************************************************/
+
+ssize_t sendv ( int socket_fd, in iovec_const[] data, int flags = 0 )
+{
+    Const!(msghdr) m = {
+            msg_iov:    cast(Const!(iovec*))data.ptr,
```
gavin-norman-sociomantic, line 137: Why is this cast needed? Afaics, `msghdr.msg_iov` is declared as a simple `iovec*` (non-const).

david-eckardt-sociomantic: Hm, good question.

david-eckardt-sociomantic: `typeof(msghdr.msg_iov)` is `iovec*` but `typeof(data.ptr)` is `iovec_const*`.

src/swarm/neo/protocol/socket/uio_const.d
david-eckardt-sociomantic: All the `sizeof`, `alignof` and `tupleof.length` checks could go in there but not easily the field type identity and equivalence checks: It verifies that the types of all fields are identical to the original struct except those that were changed to equivalent types (identical but `const` in this case). I don't see a simple way how an automatic checker could know which fields should be identical and which criterion of equivalence applies to those that should be different.

david-eckardt-sociomantic:
Updated with a description of the purpose of `iovec_const` for this particular case and `uio` vector I/O in general.

src/swarm/neo/protocol/socket/uio_const.d
gavin-norman-sociomantic: Ah, of course :D

src/swarm/neo/protocol/socket/uio_const.d
gavin-norman-sociomantic: Well, I imagined using `Unqual` on everything to compare types.

gavin-norman-sociomantic:
LGTM. I can't remember... did we discuss whether this should be applied to v2.6.x?

david-eckardt-sociomantic:
Moved to the private repository to be merged into v2.6.x.

Issue #10, opened 2017-04-28
Investigate whether suppressing of SIGPIPE can be avoided


gautam-kotian-sociomantic:
Swarm-neo suppresses the SIGPIPE signal here:
https://github.com/sociomantic/swarm/blob/2df46475b0719307706db68f20d7af72a96392b9/src/swarm/neo/protocol/socket/MessageSender.d#L57-L64

But this then requires applications which produce output that may be piped to other programs to explicitly catch SIGPIPE and exit.

gautam-kotian-sociomantic:
On slack, @david-eckardt-sociomantic hinted that this may be possible.

```
[2017-04-28 13:21] David Eckardt
a quick look reveals that we could probably replace `writev(2)` with `sendmsg(2)` there,
and `sendmsg` allows suppressing `SIGPIPE`, so we wouldn't need to ignore it
```

nemanja-boric-sociomantic:
It's already being in progress. See #9

nemanja-boric-sociomantic:
Specifically, this file: https://github.com/sociomantic-tsunami/swarm/pull/9/commits/9e36bf357f143ec49a5de0f9832f7b2b7349b233#diff-adb75301e407f1aa91acf9c69e92fed9

nemanja-boric-sociomantic:
And finally #11 :-)

gautam-kotian-sociomantic:
Good to know. Feel free to close as duplicate if that's appropriate.

nemanja-boric-sociomantic:
Fixed in https://github.com/sociomantic-tsunami/swarm/commit/fc57af9d6235174a2b7773f12aed5bf13bebb6aa

PR #11, opened 2017-04-28
Disable SIGPIPE per socket output operation rather than globally

david-eckardt-sociomantic:
Swarm neo socket output needs to suppress `SIGPIPE` for socket output while using vector output. This caused the following conflict:
- [`send(2)`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/send.html) accepts a `MSG_NOSIGNAL` flag to suppress `SIGPIPE` for output on the socket but doesn't support vector output.
- [`writev(2)`](http://pubs.opengroup.org/onlinepubs/9699919799/functions/writev.html) performs vector output but doesn't accept flags.

The solution was to ignore the signal by running `signal(SIGPIPE, SIG_IGN)` in a static constructor in `MessageSender`, which was incorrectly added in a minor release yet is a breaking change. It was reported to be a problem, and the new `sendv` function accepts `MSG_NOSIGNAL` to disable the signal just for one socket, so change `MessageSender` to use `MSG_NOSIGNAL` rather than disabling `SIGPIPE` for the whole application, fixing the bug introduced by breaking program behaviour.

### DIFF
--- a/src/swarm/neo/client/Connection.d
+++ b/src/swarm/neo/client/Connection.d
@@ -588,7 +588,7 @@ public final class Connection: ConnectionBase
     ***************************************************************************/
 
     override protected void getPayloadForSending (
-        RequestId id, void delegate ( void[][] payload ) send
+        RequestId id, void delegate ( in void[][] payload ) send
     )
     {
         if (auto request_handler = this.getRequestOnConn(id))
--- a/src/swarm/neo/client/IRequestSet.d
+++ b/src/swarm/neo/client/IRequestSet.d
@@ -59,7 +59,7 @@ interface IRequest
 
 interface IRequestOnConn
 {
-    void getPayloadForSending ( void delegate ( void[][] payload ) send );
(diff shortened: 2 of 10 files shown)


nemanja-boric-sociomantic:
Could you add "Fixes #10" at the last commit?

david-eckardt-sociomantic:
Blocked, depends on #9.

gavin-norman-sociomantic:
>This disabling SIGPIPE for the whole application, which has been reported to be a problem.

This sentence seems garbled.

david-eckardt-sociomantic:
Updated with ungarbled sentence. Should this be merged into v2.6.x like #9?

nemanja-boric-sociomantic:
I think so, since the bug is present there as well.

gavin-norman-sociomantic:
Yeah, this bug fix should be applied to v2.6.x.

david-eckardt-sociomantic:
Moved to the private repository to be merged into v2.6.x.

PR #8, opened 2017-04-27
Add simple GetAll request to test client/node

gavin-norman-sociomantic:

PR #19, opened 2017-05-12
Request event dispatcher

gavin-norman-sociomantic:
Based on #8. Blocked until merged.

### DIFF
--- a/relnotes/resume-fiber.feature.md
+++ b/relnotes/resume-fiber.feature.md
@@ -1,0 +1,8 @@
+* `swarm.neo.connection.RequestOnConnBase`
+
+  The nested class `EventDispatcher` now has the ability to resume the outer
+  class' (i.e. the request-on-conn instance) fiber. This can be useful for
+  multi-fiber requests which are communicating with each other via fiber resume
+  codes of the request-on-conn. In client request handlers, the actual
+  request-on-conn instance is not available.
+
--- a/src/swarm/neo/request/RequestEventDispatcher.d
+++ b/src/swarm/neo/request/RequestEventDispatcher.d
@@ -1,0 +1,763 @@
+/*******************************************************************************
+
+    Request event dispatcher for use with multi-fiber requests.
+
+    Three types of events are handled:
+        1. "Signals" (ubyte codes) sent between fibers.
+        2. Messages received over the connection.
+        3. Connection readiness to send a payload.
+
+    The types of events are handled somewhat differently, as follows:
+        1. Only one request handler fiber may register to be notified of each
+           type of signal.
+        2. Messages reveived over the connection are expected to begin with a
+           byte representing the message type. Only one request handler fiber
+           may register to be notified of each type of message.
+        3. Each request handler fiber may only register once (at a time) for
+           notification of connection send readiness. Send readiness
+           notifications are handled in order of registration (i.e. fibers will
+           be notified of successive send readiness in the order in which they
+           registered to receive this notification.)
+
+    Copyright: Copyright (c) 2017 sociomantic labs GmbH. All rights reserved
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+
+*******************************************************************************/
+
+module swarm.neo.request.RequestEventDispatcher;
+
+import ocean.transition;
+import ocean.core.SmartUnion;
+
+/*******************************************************************************
+
+    Smart union of events which can be registered with the dispatcher. An
+    instance of the union, registered with the dispatcher, represents a fiber
+    awaiting the event specified by the active field. A single fiber may be
+    registered for mutliple events, but some event types may only be registered
+    for a single fiber (see RequestEventDispatcher.register).
+
+*******************************************************************************/
+
+private alias SmartUnion!(EventRegistrationUnion) EventRegistration;
+
+/*******************************************************************************
+
+    Union of events which can be registered with the dispatcher.
+
+*******************************************************************************/
+
+private union EventRegistrationUnion
+{
+    /// Waiting to receive a message over the connection.
+    Message message;
+
+    /// Waiting to receive a signal from another fiber.
+    Signal signal;
+
+    /// Waiting to send a message over the connection.
+    Send send;
+}
+
+/*******************************************************************************
+
+    Struct representing a fiber awaiting a certain message over the connection.
+    Each message is required to begin with a single byte representing the
+    message type.
+
+*******************************************************************************/
+
+public struct Message
+{
+    /// The type of the message being awaited.
+    ubyte type;
+}
+
+/*******************************************************************************
+
+    Struct representing a fiber awaiting a signal with a certain code.
+
+*******************************************************************************/
+
+public struct Signal
+{
+    /// The signal being awaited.
+    ubyte code;
+}
+
+/*******************************************************************************
+
+    Struct representing a fiber awaiting the completion of sending a payload
+    over the connection.
+
+*******************************************************************************/
+
+public struct Send
+{
+    import swarm.neo.connection.RequestOnConnBase;
+
+    /// Alias for a delegate type to set the payload of the message to send.
+    alias void delegate ( RequestOnConnBase.EventDispatcher.Payload payload )
+        GetPayloadDg;
+
+    /// Delegate to set the payload of the message to send.
+    GetPayloadDg get_payload;
+}
+
+/*******************************************************************************
+
+    Smart union of events which the dispatcher can notify clients of. An
+    instance of this union is returned from the dispatcher method which
+    suspended a waiting fiber (see RequestEventDispatcher.nextEvent), informing
+    it of the event which has occurred.
+
+*******************************************************************************/
+
+public alias SmartUnion!(EventNotificationUnion) EventNotification;
+
+/*******************************************************************************
+
+    Union of events which the dispatcher can notify clients of.
+
+*******************************************************************************/
+
+private union EventNotificationUnion
+{
+    /// A message was received over the connection.
+    ReceivedMessage message;
+
+    /// A signal was passed from another fiber.
+    Signalled signal;
+
+    /// A message was sent over the connection.
+    Sent sent;
+}
+
+/*******************************************************************************
+
+    A message received over the connection, broken down into its type and
+    payload.
+
+*******************************************************************************/
+
+public struct ReceivedMessage
+{
+    /// The type of the message.
+    ubyte type;
+
+    /// The message payload.
+    Const!(void)[] payload;
+}
+
+/*******************************************************************************
+
+    A signal which occurred.
+
+*******************************************************************************/
+
+public struct Signalled
+{
+    /// Code of the signal.
+    ubyte code;
+}
+
+/*******************************************************************************
+
+    Struct representing the successful sending of a payload over the connection.
+
+*******************************************************************************/
+
+public struct Sent
+{
+    // Dummy struct with no data, used purely as a "flag" in the smart union to
+    // indicate the event which occurred. (As each fiber is only allowed to
+    // send one thing at a time (see RequestEventDispatcher.register), there is
+    // no need to for this struct to have any fields.
+}
+
+/*******************************************************************************
+
+    Request event dispatcher for use with multi-fiber requests.
+
+    Three types of events are handled:
+        1. "Signals" (ubyte codes) sent between fibers.
+        2. Messages received over the connection.
+        3. Connection readiness to send a payload.
+
+    The types of events are handled somewhat differently, as follows:
+        1. Only one request handler fiber may register to be notified of each
+           type of signal.
+        2. Messages reveived over the connection are expected to begin with a
+           byte representing the message type. Only one request handler fiber
+           may register to be notified of each type of message.
+        3. Each request handler fiber may only register once (at a time) for
+           notification of connection send readiness. Send readiness
+           notifications are handled in order of registration (i.e. fibers will
+           be notified of successive send readiness in the order in which they
+           registered to receive this notification.)
+
+*******************************************************************************/
+
+public struct RequestEventDispatcher
+{
+    import ocean.core.Enforce;
+    import Array = ocean.core.Array : moveToEnd, removeShift, contains;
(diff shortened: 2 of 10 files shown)


src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -0,0 +1,690 @@
...
+            if ( this.waitingWriters() )
+            {
+                auto writer = this.popWaitingWriter();
+                bool sent;
+
+                do
+                {
+                    this.dispatchQueuedSignals(conn);
+
+                    auto resume_code = conn.sendReceiveAndHandleEvents(
```
gavin-norman-sociomantic: I think `sendReceiveAndHandleEvents` (also `receiveAndHandleEvents`, called below) can throw, in the case of an I/O error. How should these errors be dispatched?

gavin-norman-sociomantic: Updated with handling for request-on-conn exceptions. (All waiting fibers are resumed with the exception.)

gavin-norman-sociomantic:
#8 merged. Unblocked.

gavin-norman-sociomantic:
Reviewer beware: the order of the commits displayed in github is not correct!

gavin-norman-sociomantic:
Correct commit order is:
1. Add RequestOnConnBase.EventDispatcher.resumeFiber method (https://github.com/sociomantic-tsunami/swarm/pull/19/commits/e6e4230119ab733ff01e134f80827f99d636778e)
2. Add RequestEventDispatcher helper struct (https://github.com/sociomantic-tsunami/swarm/pull/19/commits/ab53d7cd0308fb00e03850b1315045001471c62c)
3. Adapt example node's GetAll request to use worker/controller fibers (https://github.com/sociomantic-tsunami/swarm/pull/19/commits/c7c681d77b5eacddacb1e44d6a2abc89651b4066)
4. Remove commented out code (https://github.com/sociomantic-tsunami/swarm/pull/19/commits/214f5578da9cb0e2bea1546a04a710eedbc224aa)

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -0,0 +1,767 @@
...
+    Each message is required to begin with a single byte representing the
+    message type.
+
+*******************************************************************************/
+
+public struct Message
+{
+    /// The type of the message being awaited.
+    ubyte type;
+}
```
david-eckardt-sociomantic, line 76: What is the reason for single-member structs like this -- why not a `ubyte message_type` field in `EventInUnion`?

gavin-norman-sociomantic: True, could probably be simplified. (I think originally they had more fields.)

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -0,0 +1,767 @@
...
+            case send:
+                foreach ( waiting_fiber; this.waiting_fibers )
+                    if ( waiting_fiber.event.active == send )
+                        enforce(waiting_fiber.fiber != fiber,
+                            "Each fiber may only send one thing at a time");
+                break;
+            default:
+                assert(false);
+        }
+
```
david-eckardt-sociomantic, line 571: Interesting case, comparing two unions: Looks terribly redundant, but I don't see how it could be simplified.

test/neo/client/request/internal/GetAll.d
```diff
@@ -213,8 +181,178 @@ public struct GetAll
...
+            {
+                auto message = this.outer.request_event_dispatcher.receive(this.fiber,
+                    Message(MessageType.Record), Message(MessageType.End));
+                with ( MessageType ) switch ( message.type )
+                {
+                    case Record:
+                        count++;
+
+                        auto key = *conn.message_parser.
+                            getValue!(hash_t)(message.payload);
```
david-eckardt-sociomantic, line 256: Why is the key not a hexadecimal string? :trollface:

test/neo/client/request/internal/GetAll.d
```diff
@@ -213,8 +181,178 @@ public struct GetAll
...
+        void fiberMethod ( )
+        {
+            // Wait for a control message to be initiated.
+            // In a real client, there would be a user-facing API to do this. In
+            // this example, for simplicity, the Reader triggers a control
+            // message.
+            auto event = this.outer.request_event_dispatcher.nextEvent(
+                this.fiber, Signal(Signals.Stop));
+            assert(event.active == event.active.signal);
+            assert(event.signal.code == Signals.Stop);
```
david-eckardt-sociomantic: The `in` contract of `event.signal()` contains ` assert(event.active == event.active.signal)`.

src/swarm/neo/request/RequestEventDispatcher.d
gavin-norman-sociomantic: Ah actually there is a reason for this which became clear when I tried factoring it out: wrapping the union fields in structs allows for usage like this:
```D
ubyte awaited_message, awaited_signal;
RequestEventDispatcher dispatcher;
dispatcher.nextEvent(my_fiber, Message(awaited_message), Signal(awaited_signal));
```
With simple union fields, that would look like:
```D
ubyte awaited_message, awaited_signal;
EventIn message_event;
message_event.message = awaited_message;
EventIn signal_event;
signal_event.signal = awaited_signal;
RequestEventDispatcher dispatcher;
dispatcher.nextEvent(my_fiber, message_event, signal_event);
```

gavin-norman-sociomantic: (The point is that the `signal` and `message` fields of the event union are both `ubyte`. With a plain union, there's no way to distinguish them other than directly setting them by name.)

gavin-norman-sociomantic:
Updated:
* Updated to ocean v3.1.4 and removed `toString` workarounds in event unions.
* Small doc enhancements.
* Removed duplicate assertion.

gavin-norman-sociomantic:
Updated:
* Added asserts to `RequestEventDispatcher.abort` to ensure that 1. a running fiber cannot be aborted, 2. a terminated fiber can have no events registered.

src/swarm/neo/request/RequestEventDispatcher.d
david-eckardt-sociomantic: Fair enough.

relnotes/resume-fiber.feature.md
```diff
@@ -0,0 +1,8 @@
+* `swarm.neo.connection.RequestOnConnBase`
+
+  The nested class `EventDispatcher` now has the ability to suspend the outer
```
nemanja-boric-sociomantic: Strange how this talks about suspending the fiber, while this patch is adding `resumeFiber` method.

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -0,0 +1,776 @@
...
+}
+
+
+/*******************************************************************************
+
+    Smart union of events which can be registered with the dispatcher.
+
+*******************************************************************************/
+
+public alias SmartUnion!(EventInUnion) EventIn;
```
nemanja-boric-sociomantic: Where does the `In` coming from? Maybe some other name would be a better choice? I don't have any other suggestion, though.

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -0,0 +1,776 @@
...
+
+/*******************************************************************************
+
+    Union of events which can be registered with the dispatcher.
+
+*******************************************************************************/
+
+private union EventInUnion
+{
+    Message message;
```
nemanja-boric-sociomantic: Documentation.

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -0,0 +1,776 @@
...
+            msg.exc = e;
+            foreach ( waiting_fiber; this.waiting_fibers )
+                if ( waiting_fiber.fiber.waiting )
+                    waiting_fiber.fiber.resume(this.token, null, msg);
+
+            // If a writer fiber was popped from the list, abort that too.
+            if ( writer.fiber !is null && writer.fiber.waiting )
+                writer.fiber.resume(this.token, null, msg);
+
+            this.waiting_fibers.length = 0;
```
nemanja-boric-sociomantic, line 430: `enableStomping`?

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -0,0 +1,776 @@
...
+
+        Params:
+            fiber = fiber to unregister as awaiting an event
+            event = event which fiber is no longer awaiting
+
+    ***************************************************************************/
+
+    private void unregister ( MessageFiber fiber, EventIn event )
+    {
+        this.waiting_fibers.length = Array.moveToEnd(this.waiting_fibers,
```
nemanja-boric-sociomantic: `enableStomping`?

relnotes/resume-fiber.feature.md
gavin-norman-sociomantic: Ha good catch. Word mismatch.

gavin-norman-sociomantic: Updated.

src/swarm/neo/request/RequestEventDispatcher.d
gavin-norman-sociomantic: Changed to `EventRegistration` and `EventNotification`. I think that's clearer.

gavin-norman-sociomantic:
Updated.

src/swarm/neo/request/RequestEventDispatcher.d
nemanja-boric-sociomantic: Much clearer, that's what I guessed this is, thanks!

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -0,0 +1,777 @@
...
+
+    /// Fibers currently awaiting events.
+    private WaitingFiber[] waiting_fibers;
+
+    /// Event which has fired, to be returned by nextEvent(). (This value is set
+    /// by notifyWaitingFiber(), just before resuming the waiting fiber.)
+    private EventNotification last_event;
+
+    /// Token passed to fiber suspend/resume calls.
+    private static MessageFiber.Token token;
```
nemanja-boric-sociomantic: This doesn't seem to be initialized anywhere, would this be a problem? (I presume no, as IIRC, we're using `MessageFiber` here solely for its ability to kill the fiber)

gavin-norman-sociomantic: Yeah... I don't _think_ it's a problem to just use the default values. I'd like to split out the killing vs the token checking vs the message passing facilities of `MessageFiber`...

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -0,0 +1,777 @@
...
+    private void dispatchSignal ( RequestOnConnBase.EventDispatcher conn,
+        int signal )
+    {
+        if ( signal < 0 ) // Standard send/receive codes
+            return;
+
+        assert(signal <= 255);
+        ubyte signal_ubyte = cast(ubyte)signal;
+
+        foreach ( waiting_fiber; this.waiting_fibers )
```
nemanja-boric-sociomantic, line 690: One note here on the performance side (not sure if it's worth paying attention now, but it could be useful if it turns out that these loops are very tight when handling a lot of requests): Instead of a single list `this.waiting_fibers` we could have `this.fibers_waiting_for_message` and `this.fibers_waiting_for_signal` and then even these could actually be arrays of arrays, where the first dimension is the message/signal code, so you can access the first waiting fiber in constant time.

gavin-norman-sociomantic: I considered that, but remember that this is an event dispatcher _for a single request_. So we'd only need to care about this kind of optimisation if we get to the stage of having requests with dozens of fibers. If that ever happens, I think we'll have other problems to worry about too ;)

nemanja-boric-sociomantic: Ah, yes, very good point - this is an event dispatcher for a single request. All good :+1:

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -0,0 +1,777 @@
...
+            {
+                EventNotification fired_event;
+                fired_event.message = ReceivedMessage(message_type, payload);
+
+                this.notifyWaitingFiber(waiting_fiber.fiber, fired_event);
+                return;
+            }
+        }
+
+        log.warn("Unhandled message of type {}", message_type);
```
nemanja-boric-sociomantic: Last nitpick: to me it feels that this is more useful for the debugging, than it is for the live servers, and it feels in the case of the misbehaved client this could come quite costly, as there could be a big pressure on the logging system, and it could lead to the denial of service attack. I suggest either we remove this, or move it to the `info` level, so it will not be logged by default on live configurations.

gavin-norman-sociomantic: Hm, it's a tricky point. I deliberately made this a relatively high log level as it really should be noticed if this ever happens. (I'd envisaged using this in combination with the log level stats writer in ocean.)

I see what you mean, though. If the time between the buggy request starting, this error occurring, and the client firing the request again was very small, it could lead to a big performance hit affecting other requests.

nemanja-boric-sociomantic: If this is the concern, we could (in separate PR) setup stats so it logs count of how many times the connection has been shut down with the protocol error.

gavin-norman-sociomantic: You're right. That would be a better solution.

gavin-norman-sociomantic: Upated.

nemanja-boric-sociomantic (approved):
LGTM, thanks!

gavin-norman-sociomantic:
(Also updated with a non-default fiber token, just for usage correctness.)

PR #22, opened 2017-05-24
All nodes request helpers

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Based on #19 (based on #8). Blocked until parent PRs are merged.

gavin-norman-sociomantic:
Needs release notes.

gavin-norman-sociomantic:
Rebased on v4.x.x with #19 merged. Unblocked.

nemanja-boric-sociomantic:
Redish

gavin-norman-sociomantic:
Ha, that curious D1 error where you can shove template parameters on something that's not a template -- and the compiler ignores them!

nemanja-boric-sociomantic:
D1 helps you being more productive.

src/swarm/neo/client/mixins/RequestCore.d
```diff
@@ -309,7 +309,7 @@ public template RequestCore ( RequestType request_type_, ubyte request_code,
 
     ***************************************************************************/
 
-    private static void notify ( ref UserSpecifiedParams params,
+    public static void notify ( ref UserSpecifiedParams params,
```
nemanja-boric-sociomantic: release notes?

nemanja-boric-sociomantic: Documentation also says "Private helper"

nemanja-boric-sociomantic, line 313: Documentation states "private helper", also perhaps missing release notes.

gavin-norman-sociomantic:
Updated with release notes.

src/swarm/neo/client/mixins/AllNodesRequestCore.d
```diff
@@ -0,0 +1,648 @@
...
+            // Only retry in the case of a connection error. Other errors
+            // indicate internal problems and should not be retried.
+            catch ( IOError e )
+            {
+                this.disconnected(e);
+                reconnect = true;
+            }
+            finally
+            {
+                this.initialiser.reset();
```
nemanja-boric-sociomantic, line 152: Would it be a problem if the `run()` exited because `this.connector()` failed, so the `this.initialiser.initialise()` was not yet executed?

src/swarm/neo/client/mixins/AllNodesRequestCore.d
```diff
@@ -0,0 +1,648 @@
...
+    Returns:
+        always true (indicates that the request should be initialised)
+
+*******************************************************************************/
+
+public bool allNodesRequestConnector (
+    RequestOnConn.EventDispatcherAllNodes conn )
+{
+    bool connected = false;
+    do // TODO: maybe don't need a loop, if there are no other resume codes to handle
```
nemanja-boric-sociomantic: This loop is not needed, indeed. The reason is that when `waitForReconnect` returns, we get either `connected = true`, or the program aborts (see `default` clause in switch).

src/swarm/neo/client/mixins/AllNodesRequestCore.d
```diff
@@ -0,0 +1,648 @@
...
+        {
+            return allNodesRequestConnector(this.conn);
+        }
+    }
+}
+
+/*******************************************************************************
+
+    Struct template encapsulating the core logic for initialising a single
+    request-on-conn of an all-nodes request. Initialisation is deinfed as:
```
nemanja-boric-sociomantic: typo `deinfed`

src/swarm/neo/client/mixins/AllNodesRequestCore.d
```diff
@@ -0,0 +1,648 @@
...
+        Returns:
+            true if initialisation succeeded for this request-on-conn, false to
+            abort handling of the request on this connection
+
+    ***************************************************************************/
+
+    public bool initialise ( )
+    {
+        // establishConnection() should guarantee we're already connected
+        assert(this.conn.waitForReconnect() == 0);
```
nemanja-boric-sociomantic: I almost said "there shouldn't be any sideeffects inside `assert`, since it would be removed when compiling with `-release`", but this is essentially a way to say `assert(this.conn.isConnected())`. Maybe we should make that clearer (not in this PR).

src/swarm/neo/client/mixins/AllNodesRequestCore.d
```diff
@@ -0,0 +1,648 @@
...
+
+        The number of request-on-conns that are currently in the process of
+        initialising the request -- sending the initial payload and waiting for
+        a status code back from the node.
+
+        Used to decide when to send the started notification to the user.
+
+    ***************************************************************************/
+
+    public uint initialising;
```
nemanja-boric-sociomantic: I wonder if we make all this field `private`, would that make compiler enforce that it's touched only by the other entities in this module. This would enforce that the user defined code doesn't change this field.

src/swarm/neo/client/mixins/AllNodesRequestCore.d
```diff
@@ -0,0 +1,648 @@
...
+    }
+
+    /***************************************************************************
+
+        Called by AllNodesRequest to reset any state required for initialisation
+        to begin again. (In this case, there is no state that needs resetting.)
+
+    ***************************************************************************/
+
+    public void reset ( )
```
nemanja-boric-sociomantic: Nothing to do with the `all_nodes` in case of reset?

src/swarm/neo/client/mixins/AllNodesRequestCore.d
david-eckardt-sociomantic: :+1:
`waitForReconnect` seems to potentially have a side effect (that is, to wait), and IMHO assertions should not have any side effects regardless of whether code relies on them or not.

gavin-norman-sociomantic: Ah, you're right. In the case where the assertion will _fail_, there could be side-effects. That's obviously not what I intended. Yeah, `assert(this.conn.isConnected())` is what I intended. I'll have a look at how to get that information.

src/swarm/neo/client/mixins/AllNodesRequestCore.d
gavin-norman-sociomantic: Right, it was just an old note to myself that I forgot to address before pushing.

src/swarm/neo/client/mixins/AllNodesRequestCore.d
gavin-norman-sociomantic: I don't think so. `initialiser.reset` should just reset the initialiser to its initial, default state. If `initialiser.inisialise` hasn't been called, it should be in its default state anyway.

Did you have anything specific in mind?

src/swarm/neo/client/mixins/AllNodesRequestCore.d
gavin-norman-sociomantic: Updated with an extra commit adding `client.RequestOnConn.AllNodesEventDispatcher.connection_status`.

src/swarm/neo/client/mixins/AllNodesRequestCore.d
gavin-norman-sociomantic: In this PR, it can be. It's read externally in a subsequent PR, but I'll change it to private here and add a getter in the other PR.

src/swarm/neo/client/mixins/AllNodesRequestCore.d
gavin-norman-sociomantic: Nope. `AllNodesRequestSharedWorkingData.initialising` is reset in the `scope(exit)` of `initialise` and `AllNodesRequestSharedWorkingData.called_started_notifier` should never be reset because it must be called strictly once ever during the lifetime of the request.

gavin-norman-sociomantic:
Updated:
* Removed TODO / loop.
* Added connection status getter, use that in assert.
* Made `AllNodesRequestSharedWorkingData.initialising` private.
* Fixed typo.

src/swarm/neo/client/mixins/AllNodesRequestCore.d
gavin-norman-sociomantic: I've added this explanation to the method doc.

src/swarm/neo/client/mixins/AllNodesRequestCore.d
nemanja-boric-sociomantic: Nope, that's all clarification I need. Thanks!

src/swarm/neo/client/mixins/AllNodesRequestCore.d
nemanja-boric-sociomantic: :+1: I wasn't aware of the next PR.

nemanja-boric-sociomantic (approved):

PR #23, opened 2017-05-26
Add gitignore and neptune.yml

mathias-lang-sociomantic:
Those files where in the private repository, but didn't make it to the open-source one.

CC @mihails-strasuns-sociomantic : Tsunami swarm is not part of the neptune overview.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+build/
+Config.local.mak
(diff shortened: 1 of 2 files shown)


mihails-strasuns-sociomantic (approved):

mihails-strasuns-sociomantic:
Good catch!

PR #33, opened 2017-06-08
Fix compilation of ClientCore when RequestStats not instantiated

gavin-norman-sociomantic:
(Bug fix for unreleased code in v4.x.x.)

The method logRequestStats assumed that a class called RequestStats exists.
This is not always the case. It's safer to put the logging method inside
RequestStatsTemplate, so it can only be called when the template is
instantiated.


nemanja-boric-sociomantic:
Ah, yes, good point.

nemanja-boric-sociomantic (approved):

gavin-norman-sociomantic:
I only noticed when I tried compiling a client that doesn't (yet) use the request stats template with swarm v4.x.x :)

PR #35, opened 2017-06-09
Fix method intended for public use being private

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Oops :blush:

nemanja-boric-sociomantic:
Haha :-). Partial blame goes on me, I did ask the question about the visibility of this method, just it was the wrong question (the code where this is mixed in will be able to use it, but not the user code!).

nemanja-boric-sociomantic (approved):

gavin-norman-sociomantic:
Yeah, one of those last minute changes to a PR which then isn't properly tested.

PR #36, opened 2017-06-21
Add missing code to reset RequestEventDispatchers when used in a pool

gavin-norman-sociomantic:


gavin-norman-sociomantic:
I was originally viewing this as a bug fix for v4.1.x, so didn't include release notes. Then I thought it's more like a feature PR. What do people think?

nemanja-boric-sociomantic:
I think this should go as a bugfix. Otherwise, no real implementation is able to use the facility, which is included in v4.1.x, and that was not our intention.

gavin-norman-sociomantic:
Yeah that was my initial thought, until I noticed that it's actually just pure new features. I'll rebase it.

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -235,6 +235,24 @@ public struct RequestEventDispatcher
 
     /***************************************************************************
 
+        Resets this instance to its initial state. This method should be called
+        when retriving an instance from a pool or free list.
```
nemanja-boric-sociomantic: typo `retriving`

gavin-norman-sociomantic: Updated.

nemanja-boric-sociomantic (approved):

PR #43, opened 2017-07-04
Add unified method of waiting for a next event on a RoC

### LINKED ISSUE #41: Unify RequestOnConn.EventDispatcher methods for sending / receiving / yielding
Currently there are only methods to:
1. Yield and handle events.
2. Yield, receive, and handle events.

We also need:
3. Yield, send, and handle events.
4. Yield, send, receive, and handle events.

The combinations are getting silly. All these methods should be refactored into a single method which waits for one of the specified types of events and returns which event occurred.

gavin-norman-sociomantic:
Fixes #41.

### DIFF
--- a/relnotes/roc-event-dispatcher-next-event.feature.md
+++ b/relnotes/roc-event-dispatcher-next-event.feature.md
@@ -1,0 +1,8 @@
+* `swarm.neo.connection.RequestOnConnBase`
+
+  A new method -- `nextEvent` -- has been added to
+  `RequestOnConnBase.EventDispatcher`. This is intended to (eventually) replace
+  the myriad of methods providing handling of {reads/sends/yields/misc fiber
+  resume codes} in (almost) all possible combinations (an approach which is
+  clearly not maintainable).
+
--- a/src/swarm/neo/connection/RequestOnConnBase.d
+++ b/src/swarm/neo/connection/RequestOnConnBase.d
@@ -228,6 +228,7 @@ abstract class RequestOnConnBase
         import swarm.neo.IPAddress;
         import swarm.neo.protocol.MessageParser;
         import ocean.core.Traits: hasIndirections, StripEnum;
+        import ocean.core.SmartUnion;
 
         alias RequestOnConnBase.FiberResumeCode FiberResumeCode;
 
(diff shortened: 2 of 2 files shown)


src/swarm/neo/connection/RequestOnConnBase.d
```diff
@@ -507,6 +614,183 @@ abstract class RequestOnConnBase
...
+
+            Returns:
+                an EventNotification instance denoting the event which occurred
+
+            Throws:
+                Exception on protocol or I/O error.
+
+        ***********************************************************************/
+
+        public EventNotification nextEvent (
```
gavin-norman-sociomantic, line 645: I'm interested whether anyone can think of a nicer API than this (the arguments). I'm not super keen on it, as it stands.

gavin-norman-sociomantic:
Fixed D2 compile error.

src/swarm/neo/connection/RequestOnConnBase.d
```diff
@@ -243,6 +244,112 @@ abstract class RequestOnConnBase
...
 
+            Bitfield enum to specify the events to wait on. Passed to the
+            nextEvent() method.
+
+        ***********************************************************************/
+
+        public enum Flags
+        {
+            /// Wait to receive a payload for this request over the connection.
+            Receive     = 0b01,
```
nemanja-boric-sociomantic: Nitpick: I tend to prefer `1 << 0`, `1 << 1`, etc.

src/swarm/neo/connection/RequestOnConnBase.d
```diff
@@ -243,6 +244,112 @@ abstract class RequestOnConnBase
...
+
+        /// Type of a delegate to fill in a payload for sending.
+        alias void delegate ( RequestOnConnBase.EventDispatcher.Payload payload )
+            FillPayloadDg;
+
+        /***********************************************************************
+
+            Smart union of events which the dispatcher can notify clients of. An
+            instance of this union is returned from the dispatcher method which
+            suspends a waiting fiber (see nextEvent()), informing the called of
```
nemanja-boric-sociomantic: `informing the called`?

src/swarm/neo/connection/RequestOnConnBase.d
```diff
@@ -243,6 +244,112 @@ abstract class RequestOnConnBase
 
         /***********************************************************************
 
+            Bitfield enum to specify the events to wait on. Passed to the
+            nextEvent() method.
+
+        ***********************************************************************/
+
+        public enum Flags
```
nemanja-boric-sociomantic: I would name this enum differently, since it's now a member of `EventDispatcher`, but it's really only relevant for `nextEvent`.

src/swarm/neo/connection/RequestOnConnBase.d
nemanja-boric-sociomantic: Perhaps @david-eckardt-sociomantic would have a better idea, I don't have any :(.

src/swarm/neo/connection/RequestOnConnBase.d
```diff
@@ -507,6 +614,185 @@ abstract class RequestOnConnBase
...
+
+            scope ( exit )
+            {
+                if ( receiving )
+                    this.outer.connection.unregisterForErrorNotification(
+                        this.outer.request_id);
+            }
+
+            // Set up yielding, if required.
+            scope yielded_rqonconn = this.outer.new YieldedRequestOnConn;
```
nemanja-boric-sociomantic, line 719: Too bad there's no nice way to escape creating `yielded_rqonconn` and `payload` even if you don't need them.

src/swarm/neo/connection/RequestOnConnBase.d
```diff
@@ -507,6 +614,185 @@ abstract class RequestOnConnBase
...
+                    if ( sending )
+                    {
+                        fired_event.sent = Sent();
+                        sending = false; // don't unregister on exit; already done
+                    }
+                    else
+                    {
+                        auto e = this.protocol_error.set(
+                            "RequestOnConn unexpectedly resumed with Sent");
+                        this.shutdownConnection(e);
```
nemanja-boric-sociomantic, line 747: This throws in all other RoC's and then you throw in this one, right?

nemanja-boric-sociomantic (commented):
Left some minor comments. Although it's convoluted, it makes sense what this method is doing after reading it - it's not that bad.

API feels leaking details (like `enum Flags` to `EventDispatcher`), but I don't know what else could be done

src/swarm/neo/connection/RequestOnConnBase.d
gavin-norman-sociomantic: Good point.

src/swarm/neo/connection/RequestOnConnBase.d
gavin-norman-sociomantic: True, that is slightly nicer.

src/swarm/neo/connection/RequestOnConnBase.d
gavin-norman-sociomantic: I meant "caller"

src/swarm/neo/connection/RequestOnConnBase.d
gavin-norman-sociomantic: Yeah... I can't see a way around that, unfortunately.

src/swarm/neo/connection/RequestOnConnBase.d
gavin-norman-sociomantic: I think so. To be honest, I just copied this from the old functions, so I've not re-checked exactly what it does.

david-eckardt-sociomantic: This is what it does.

src/swarm/neo/connection/RequestOnConnBase.d
gavin-norman-sociomantic: Something like this should work and looks cleaner.
```D
auto fired_event = this.connection.waitForNextEvent!("yield", "receive", &myDelegate)();
```

nemanja-boric-sociomantic: Would passing the delegate as a template parameter work?

gavin-norman-sociomantic: Have to try it. Probably as an alias or something.

nemanja-boric-sociomantic: Yeah, better try it. This "can't have two contexts" thing is floating around in my head. If worst comes to worst, we can always pass just that one as a runtime arg.

gavin-norman-sociomantic: Exactly.

nemanja-boric-sociomantic:
Are we going to change the API to a string literal template args?

gavin-norman-sociomantic:
I'm just trying it out.

nemanja-boric-sociomantic:
OK, then I'll wait for merging this.

gavin-norman-sociomantic:
Updated.

relnotes/roc-event-dispatcher-next-event.feature.md
```diff
@@ -0,0 +1,8 @@
+* `swarm.neo.connection.RequestOnConnBase`
+
+  A new method -- `nextEvent` -- has been added to
+  `RequestOnConnBase.EventDispatcher`. This is intended to (eventually) replace
+  the myriad of methods providing handling of {reads/sends/yields/misc fiber
```
nemanja-boric-sociomantic, line 5: Just one question: what are these curly brackets for?

nemanja-boric-sociomantic (approved):

relnotes/roc-event-dispatcher-next-event.feature.md
gavin-norman-sociomantic: My attempt to represent an unordered list of possibilities.

PR #45, opened 2017-07-06
Enable socket output buffering and enable request handlers to flush the buffer

david-eckardt-sociomantic:
Enable buffering outgoing data on the socket via TCP Cork and add `RequestOnConnBase.flush()` for request handlers to flush the buffer if needed.

For the record, TCP Cork  is a socket option an can be enabled or disabled for a socket via `setsockopt`. If enabled, outgoing data are aggregated in a buffer until either a full size TCP frame can be sent or 200ms have elapsed or TCP Cork has been disabled for the socket. This means that the output buffer can be deliberately flushed by disabling and then re-enabling `TCP_CORK` for a socket. Reference: [`man 7 tcp`](http://man7.org/linux/man-pages/man7/tcp.7.html).

Control messages such as "Suspend" or "Stop" are small (tens of bytes) and therefore are likely held back by TCP Cork, but they need to be passed on as soon as possible. So a request handler needs a way of flushing the TCP Cork buffer: `RequestOnConnBase.EventDispatcher.flush()`, which calls `MessageSender.flush()`.
Of course this is useful only if `flush()` is called right after the socket `write` call -- *right after* meaning that no event loop cycle has been done or other request or I/O handler fibers have been active in the mean time. This is accomplished if the request handler calls `flush()` right after one of the `EventDispatcher.send*` methods has returned, which results in the following program flow:
`EventDispatcher.send*` either is or calls one of the `send(Receive)AndHandleEvents` methods, which calls `RequestOnConnBase.registerForSending`. Now there are two possibilities.
1. The send loop is idle.
    - `SendLoop.registerForSending` resumes the send fiber (from the request handler fiber).
    - `SendLoop.loop` calls `SendLoop.sendRequestPayload`, which calls `RequestOnConnBase.getPayloadForSending`.
    - `RequestOnConnBase.getPayloadForSending` calls the `send` callback, which eventually makes the socket `write` call, and returns. It notices that the request handler fiber is not suspended so it doesn't resume it and returns right away.
   - Depending on `finish_sending`
        - if `false`, `SendLoop.sendRequestPayload` returns and `SendLoop.loop` suspends the send fiber (since the send loop was idle and no other request handler fiber has been active the send queue is now empty);
        - if `true`, calling `MessageSender.finishSending` suspends the send fiber.
    Either way the send fiber is suspended and the request handler fiber continues.
    - `RequestOnConnBase.registerForSending` returns.
    -  `send(Receive)AndHandleEvents` sees that `outer.send_payload_ is null` and returns.
    - The request now calls `EventDispatcher.flush`, which is right after the socket `write` call has been made.
2. The send loop is busy.
    - `SendLoop.registerForSending` pushes the request id in the send queue and returns.
    - `send(Receive)AndHandleEvents` sees that `outer.send_payload_` is still there and suspends the request handler fiber.
    - At some point the send fiber is active, `SendLoop.loop` pops the request id from the send queue and calls `SendLoop.sendRequestPayload`, which calls `RequestOnConnBase.getPayloadForSending`.
    - `RequestOnConnBase.getPayloadForSending` calls the `send` callback, which eventually makes the socket `write` call, and returns. It notices that the request handler fiber is suspended so it resumes it.
    - `send(Receive)AndHandleEvents` returns.
    - The request now calls `EventDispatcher.flush`, which is right after the socket `write` call has been made.

### DIFF
--- a/src/swarm/neo/connection/ConnectionBase.d
+++ b/src/swarm/neo/connection/ConnectionBase.d
@@ -105,6 +105,16 @@ abstract class ConnectionBase: ISelectClient
 
         /***********************************************************************
 
+            If this flag is true and `sendRequestPayload` is waiting for
+            `EPOLLOUT` then it will call `sender.flush` after it has sent the
+            remaining data.
+
+        ***********************************************************************/
+
+        public bool flush_requested;
+
+        /***********************************************************************
+
             The queue of ids of requests waiting to send a message.
 
         ***********************************************************************/
@@ -353,6 +363,8 @@ abstract class ConnectionBase: ISelectClient
         {
             bool finish_sending = false;
 
+            scope (exit) this.flush_requested = false;
+
             this.outer.getPayloadForSending(
                 id,
                 ( in void[][] payload )
@@ -369,6 +381,8 @@ abstract class ConnectionBase: ISelectClient
                 // Register the socket for both input and output
                 this.outer.registerEpoll(true);
                 this.outer.sender.finishSending(this.suspend());
+                if (this.flush_requested)
+                    this.outer.sender.flush();
             }
         }
     }
@@ -1056,6 +1070,31 @@ abstract class ConnectionBase: ISelectClient
 
     abstract protected void getPayloadForSending ( RequestId id,
         void delegate ( in void[][] payload ) send );
+
+    /***************************************************************************
+
+        Requests to send outgoing data buffered by the OS for the socket
+        immediately. By default outgoing data are buffered to the maximum
+        payload of a TCP frame using the Linux TCP_CORK socket option.
+
+        This method is intended to be called right after `getPayloadForSending`
+        has called the `send` delegate.
+
+    ***************************************************************************/
+
+    public void flush ( )
+    {
+        // Flush the OS-maintained (TCP_CORK) socket output buffer. This will
+        // immediately send all data that have been accepted by `sendmsg(2)`.
+        this.sender.flush();
+        // `this.sender` may in addition buffer data that have not been accepted
+        // by `sendmsg` - that is, `send(2)` returned a value less than the
+        // buffer length argument - and wait for `EPOLLOUT` on the socket to
+        // call `sendmsg` again  with the remaining data. Set a flag to flush
+        // the OS-maintained (TCP_CORK) socket output buffer when `sendmsg`
+        // eventually accepted all outstanding data.
+        this.send_loop.flush_requested = true;
+    }
 
     /***************************************************************************
 
--- a/src/swarm/neo/connection/ConnectionBase.d
+++ b/src/swarm/neo/connection/ConnectionBase.d
@@ -262,6 +262,8 @@ abstract class ConnectionBase: ISelectClient
 
                 try
                 {
+                    this.outer.sender.cork = true;
+
                     // Start the receive fiber, it will suspend itself immediately.
                     this.outer.recv_loop.start();
 
--- a/src/swarm/neo/connection/RequestOnConnBase.d
+++ b/src/swarm/neo/connection/RequestOnConnBase.d
@@ -482,6 +482,19 @@ abstract class RequestOnConnBase
 
         /***********************************************************************
 
+            Requests to send outgoing data buffered by the OS for the socket
+            immediately. By default outgoing data are buffered to the maximum
+            payload of a TCP frame using the Linux TCP_CORK socket option.
+
+        ***********************************************************************/
+
+        public void flush ( )
+        {
+            this.outer.connection.flush();
+        }
+
+        /***********************************************************************
+
             Sends a message for this request to the node.
 
             Do not resume the fiber before this method has returned or thrown.


src/swarm/neo/connection/ConnectionBase.d
```diff
@@ -1034,6 +1048,23 @@ abstract class ConnectionBase: ISelectClient
...
+        payload of a TCP frame using the Linux TCP_CORK socket option.
+
+        This method is intended to be called right after `getPayloadForSending`
+        has called the `send` delegate.
+
+    ***************************************************************************/
+
+    public void flush ( )
+    {
+        this.sender.flush();
```
gavin-norman-sociomantic: Why is `sender.flush` called here _and_ at line 389, above?

src/swarm/neo/connection/ConnectionBase.d
```diff
@@ -1034,6 +1048,23 @@ abstract class ConnectionBase: ISelectClient
 
     /***************************************************************************
 
+        Requests to send outgoing data buffered by the OS for the socket
+        immediately. By default outgoing data are buffered to the maximum
+        payload of a TCP frame using the Linux TCP_CORK socket option.
+
+        This method is intended to be called right after `getPayloadForSending`
```
gavin-norman-sociomantic, line 1082: I can imagine an alternate approach, where the send delegate specifies (via a return value or out parameter) whether flushing is desired or not.

I'd be interested to see what the usage of both cases looks like.

What do you think?

gavin-norman-sociomantic (commented):
Two additional points:
1. We'll also need a way of flushing writes triggered by the `RequestEventDispatcher`. We can talk about this.
2. It'd be nice to adapt one of the test requests (`swarm.test.neo`) to use flush, to confirm that it doesn't alter request behaviour. The test GetAll request, which uses the controller pattern, is probably the best candidate. This will require 1.

gavin-norman-sociomantic:
Hm, interesting, the neo test hangs. I wonder if this is a problem with the build system or actually that this change has affected the test behaviour?

gavin-norman-sociomantic:
This is on the edge between a feature and a bug-fix, I feel. It's a feature that was supposed to exist in a previous release but which wasn't fully implemented.

It would certainly be more convenient, in terms of updating applications to use this fix, if it were included in a patch release, not a minor release ;)

src/swarm/neo/connection/ConnectionBase.d
nemanja-boric-sociomantic: This is what @david-eckardt-sociomantic and I initially talked about but we were worried about the explosion of the overload that we would need to support (no breaking changes, so we would have to introduce an overload for each `send` method).

src/swarm/neo/connection/ConnectionBase.d
nemanja-boric-sociomantic: I agree this needs to be better documented, but the point is following: 

There are two possibilities:

1) You write something to socket, write succeeds, you uncork the socket and you're done. This is what this `this.sender.flush` is covering.
2) You didn't manage to send anything, but you will eventually, after the line 387. Now you have data written to your socket buffer, and you want to get it out - you need to flush it at this point. 

One could argue that, if 2. is true, you don't need to call flush at this point, just at the second, but I don't think that's possible with this API - you need somehow to pass this info to `sendRequestPayload` (probably via `out` parameter or return value, as you already suggested).

src/swarm/neo/connection/ConnectionBase.d
david-eckardt-sociomantic: We thought about a `bool` return value for `getPayloadForSending` to request flushing. This is problematic because[`RequestOnConnBase.getPayloadForSending`](https://github.com/sociomantic-tsunami/swarm/blob/v3.x.x/src/swarm/neo/connection/RequestOnConnBase.d#L1394) resumes the request handler fiber before it returns. So first the request moves on and does its business, and only when it suspends the next time `getPayloadForSending` would return, possibly significantly later.

Another idea is a `bool flush` parameter in the `send` delegate. This would work, but the parameter would need to be passed to all `RequestOnConnBase.send*` methods and stored in a `RequestOnConnBase` class variable so that `getPayloadForSending` can pass it to the `send` delegate. Considering the existing convolution of the `send/Receive/andHandleEvents)` methods this is practicable only with the redesigned one-for-all method.

gavin-norman-sociomantic: Ok, maybe it's something we can add to the all-in-one method, when that's merged.

david-eckardt-sociomantic:
The tests failed because of a race condition fixed in #46.
Rebased on for patch release. Depends on #46, blocked.

gavin-norman-sociomantic:
#46 merged. Can be rebased and unblocked.

david-eckardt-sociomantic:
Rebased and unblocked.

src/swarm/neo/connection/ConnectionBase.d
gavin-norman-sociomantic: Ok. @david-eckardt-sociomantic could you extend the doc a bit to explain the reason for these two cases?

test/neo/client/request/internal/GetAll.d
```diff
@@ -433,6 +435,8 @@ private struct Controller
         controller.handle(this.conn, this.context,
             this.request_event_dispatcher, this.fiber);
 
+        this.conn.flush();
```
gavin-norman-sociomantic: Actually, this flush should be added here https://github.com/sociomantic-tsunami/swarm/blob/v4.x.x/src/swarm/neo/client/mixins/SuspendableRequestCore.d#L814

gavin-norman-sociomantic:
In the last commit ("Call flush() in neo test") message, could you add a short explanation as to why the flushes are added in those locations?

gavin-norman-sociomantic:
Regarding adding flushes, is that strictly necessary for the fix (i.e. does the test fail without them)? I wonder if this change is more suitable for a separate PR, in a minor release.

gavin-norman-sociomantic:
This should probably be part of [v3.3.1](https://github.com/sociomantic-tsunami/swarm/milestone/11), actually.

david-eckardt-sociomantic:
Rebased for v3.3.1. Removed calling `flush` in tests, that will need to go in a new PR after v3.3.1 has been merged into v4.

PR #46, opened 2017-07-07
GetAllTest: Fix race condition between node End and client control message

david-eckardt-sociomantic:
After the node sent the End message to the client it may receive a control message which the client sent before it received or processed the `End` message. Handle this as follows: The client needs to `Ack`nowledge the `End` message. The node ends the request after receiving that `Ack` and ignores incoming control messages in the mean time.

### DIFF
--- a/test/neo/client/request/internal/GetAll.d
+++ b/test/neo/client/request/internal/GetAll.d
@@ -399,6 +399,14 @@ private struct Reader
         }
         while ( !finished );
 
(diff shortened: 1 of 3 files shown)


david-eckardt-sociomantic:
Ignore the failed continuous-integration/travis-ci/push test, I accidentally pushed to the blessed repository.

david-eckardt-sociomantic:
This bug was found in #45.

gavin-norman-sociomantic (approved):

PR #49, opened 2017-07-11
Add LibFeatures module

gavin-norman-sociomantic:
Meta module defining constants that match swarm feature releases
allowing downstream libraries to conditionally implement
functionality relying on later swarm versions.

### DIFF
--- a/src/swarm/LibFeatures.d
+++ b/src/swarm/LibFeatures.d
@@ -1,0 +1,23 @@
+/*******************************************************************************
+
+    Meta module defining constants that match swarm feature releases allowing
+    downstream libraries to conditionally implement functionality relying on
+    later swarm versions.
+
+    Copyright:
+        Copyright (c) 2017 sociomantic labs GmbH. All rights reserved
+
+    License:
+        Boost Software License Version 1.0. See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module swarm.LibFeatures;
+
+const has_features_4_2 = true;
+const has_features_4_1 = true;
+const has_features_4_0 = true;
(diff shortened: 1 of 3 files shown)


nemanja-boric-sociomantic:
Is there any point in including this file earlier? (I mean, to go via patch release)

gavin-norman-sociomantic:
Ha, we could have released it in v3.3.0 :D

nemanja-boric-sociomantic:
Well, I mean - is there any practical point of getting this into a patch release, meaning - will the downstream user benefit from providing this file earlier. If not, I think it's fine to include it in v4.2.0.

mihails-strasuns-sociomantic:
Well I think we can be pretty sure that only possible users of this functionality are you yourselves (in protos) so unless there is a use case calling for earlier inclusion now, 4.2.0 is OK

src/swarm/LibFeatures.d
```diff
@@ -0,0 +1,23 @@
...
+        Copyright (c) 2017 sociomantic labs GmbH. All rights reserved
+
+    License:
+        Boost Software License Version 1.0. See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module swarm.LibFeatures;
+
+const has_features_4_2 = true;
```
mihails-strasuns-sociomantic, line 17: Maybe sanity check test in `LibFeatures_test`? :) i.e. `static assert ( hasFeaturesFrom!("swarm", 4, 0));`

gavin-norman-sociomantic: Updated.

mihails-strasuns-sociomantic (approved):

PR #54, opened 2017-07-11
Flush after sending where needed

david-eckardt-sociomantic:
- Flush after sending a control message because control messages need to be delivered with low latency. In swarm control messages are semt in suspendable requests. (Are there more places in swarm? I didn't find any.)
- In the neo test flush after sending a control or Get/Put message. The test does one Get/Put request at a time so without flushing it takes too long.


nemanja-boric-sociomantic:
> Flush after sending a control message because control messages need to be delivered with low latency. In swarm control messages are semt in suspendable requests. (Are there more places in swarm? I didn't find any.)

No, there are no more as far as I know.

PR #73, opened 2017-07-18
Enable TCP_KEEPALIVE after the connection with the remote party

nemanja-boric-sociomantic:
TCP_KEEPALIVE socket option should be set on the TCP connection
represented by socket, not on the socket that doesn't back any
connection. This enables it only after the connection has been made, not
in the constructor which doesn't need to have the connected socket.

### DIFF
--- a/src/swarm/neo/client/Connection.d
+++ b/src/swarm/neo/client/Connection.d
@@ -496,6 +496,7 @@ public final class Connection: ConnectionBase
                 this.client_socket, this.node_address, this.send_loop,
                 this.epoll, this.receiver, this.sender, this.protocol_error
             );
+            enableKeepAlive(this.client_socket.socket);
 
             debug ( SwarmConn )
                 Stdout.formatln("{}:{}: Connection.tryConnect() succeeded",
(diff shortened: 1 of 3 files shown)


src/swarm/neo/connection/ConnectionBase.d
```diff
@@ -972,26 +971,26 @@ abstract class ConnectionBase: ISelectClient
```
gavin-norman-sociomantic, line 972: Params

gavin-norman-sociomantic (commented):
Mini comment

src/swarm/neo/client/Connection.d
```diff
@@ -496,6 +496,7 @@ public final class Connection: ConnectionBase
                 this.client_socket, this.node_address, this.send_loop,
                 this.epoll, this.receiver, this.sender, this.protocol_error
             );
+            this.enableKeepAlive(this.client_socket.socket);
```
mathias-lang-sociomantic: It's quite misleading to use `this` for `static` functions though

src/swarm/neo/connection/ConnectionBase.d
nemanja-boric-sociomantic: Thanks. Will add it. The thing is we were not checking the return value of the setsock anyway.

nemanja-boric-sociomantic:
Updated with the suggestion from @mathias-lang-sociomantic and from @gavin-norman-sociomantic and also with the error checking inside the `enableKeepAlive`.

david-eckardt-sociomantic (approved):

PR #74, opened 2017-07-19
Don't use deprecated Formatter.sformat overload if the replacement is…

nemanja-boric-sociomantic:
… present

Ocean v2.8.x+ deprecates sformat's overload that accepts a sink
returning size_t in favour of the one returning void. In order to
suppress deprecation messages, swarm wraps the sformat and chooses the
right overload based on the ocean's version check.

### DIFF
--- a/src/swarm/neo/util/Formatter.d
+++ b/src/swarm/neo/util/Formatter.d
@@ -1,0 +1,77 @@
+/*******************************************************************************
+
+    Wrapper around Ocean's ocean.text.convert.Formatter.sformat used to
+    avoid usage of the deprecated overload that accepts a sink returning
+    size_t in case overload accepting sink returning void is present (since
+    ocean v2.8.
+
+    This module could be removed and all code reverted to simply use
+    ocean's Formatter in the next major.
+
+    Copyright: Copyright (c) 2016-2017 sociomantic labs GmbH. All rights reserved
+
+    License:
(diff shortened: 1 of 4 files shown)


mihails-strasuns-sociomantic (approved):
Very nice, have you checked it with any project that was reported to suffer from deprecation madness? :)

nemanja-boric-sociomantic:
I've checked with the unittests of swarm. I will check with the project soon as well.

nemanja-boric-sociomantic:
Tested with the downstream project, no deprecations.

src/swarm/neo/util/Formatter.d
```diff
@@ -0,0 +1,74 @@
+/*******************************************************************************
+
+    Wrapper around Ocean's ocean.text.convert.Formatter.sformat used to
+    avoid usage of the deprecated overload that accepts a sink returning
+    size_t in case overload accepting sink returning void is present (since
+    ocean v2.8.
+
```
gavin-norman-sociomantic, line 7: Maybe a note that this module could be removed and revert all code to simply using ocean's `Formatter`, in the next major?

gavin-norman-sociomantic (commented):
:+1:

nemanja-boric-sociomantic:
Updated with the note about removing the module in the next major.

PR #77, opened 2017-07-19
Fix bug with signal firing multiple times in RequestEventDispatcher

gavin-norman-sociomantic:
Each signal is only allowed to be handled by one registered
fiber, so we must only add each signal once to the list -- the
waiting fiber doesn't care how *often* the signal has fired.
(Otherwise, a signal can be popped from the list, the waiting
fiber resumed, then the signal popped again -- now there's no
fiber waiting for it.)


nemanja-boric-sociomantic (approved):

nemanja-boric-sociomantic (approved):

nemanja-boric-sociomantic:
Looks like GitHub allows double approvals.

PR #82, opened 2017-07-20
Test loading client auth name/key from a file

### LINKED ISSUE #76: Add test for reading client auth from a file
Combines #53 and #47.

gavin-norman-sociomantic:
Tests the constructor added in 00a171f48826c84a27093f1a2928b7b472cb3a3b.

Fixes #76.

### DIFF
--- a/test/client_auth/main.d
+++ b/test/client_auth/main.d
@@ -1,0 +1,62 @@
+/*******************************************************************************
+
+    Test for loading client auth name/key from a file.
+
+    Copyright:
+        Copyright (c) 2017 sociomantic labs GmbH. All rights reserved
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+
+*******************************************************************************/
+
+module test.client_auth.main;
+
+import test.neo.client.Client;
+
+import swarm.neo.authentication.CredentialsFile;
+
+import ocean.transition;
+import ocean.util.test.DirectorySandbox;
+import ocean.core.Test;
+import ocean.io.device.File;
+import ocean.io.select.EpollSelectDispatcher;
+
+void main ( )
+{
+    // Create temporary sandbox directory to write files to.
+    auto sandbox = DirectorySandbox.create();
+    scope (exit) sandbox.remove();
+
+    // Helper function to write a credentials file and instantiate a client to
+    // read it.
+    void loadCredentials ( cstring credentials )
+    {
+        auto file = new File("credentials", File.ReadWriteCreate);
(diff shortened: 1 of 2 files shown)


test/client_auth/main.d
```diff
@@ -0,0 +1,86 @@
...
+void main ( )
+{
+    // Exception to throw on failure of C functions.
+    auto e = new ErrnoException;
+
+    // Get current working dir.
+    char[4096] old_cwd;
+    e.enforceRetPtr!(getcwd).call(old_cwd.ptr, old_cwd.length);
+
+    // Create a temp dir.
```
nemanja-boric-sociomantic: You could use `DirectorySandbox` utility from the ocean: https://github.com/sociomantic-tsunami/ocean/blob/v2.x.x/src/ocean/util/test/DirectorySandbox.d

```
auto sandbox = DirectorySandbox.create();
scope (exit) sandbox.remove();
```

nemanja-boric-sociomantic: (this will also chdir)

test/client_auth/main.d
```diff
@@ -0,0 +1,86 @@
...
+    void loadCredentials ( cstring credentials )
+    {
+        auto file = new File("credentials", File.ReadWriteCreate);
+        file.write(credentials);
+        file.close();
+
+        void connNotifier ( Client.Neo.ConnNotification ) { }
+
+        auto client = new Client(new EpollSelectDispatcher, "credentials",
+            "127.0.0.1", 10_000, &connNotifier);
```
david-eckardt-sociomantic, line 42: You could just create a temporary credentials file here with
```d
mstring credentialsfilename = "/tmp/Dunittest-credentials-XXXXXX\0".dup[0 .. $ - 1];
int fd = e.enforceRet!(mkstemp)(function (int fd) {return fd >= 0;}).call(credentialsfilename.ptr);
scope (exit)
{
    close(fd);
    e.enforceRetCode!(unlink).call(credentialsfilename.ptr);
}
// Write credentials to the file...
// Pass the credentials file name to the client.
auto client = new Client(new EpollSelectDispatcher, credentialsfilename,
    "127.0.0.1", 10_000, &connNotifier);
```
This would avoid changing the directory.

test/client_auth/main.d
gavin-norman-sociomantic: Ahh I had a vague feeling there as something like that. Should have looked.

gavin-norman-sociomantic:
Updated to use sandbox helper from ocean.

Issue #80, opened 2017-07-20
It must be possible to disable output buffering via TCP cork


gavin-norman-sociomantic:
dd6c928467c79cd17fa17d870d2fdbc0d9093916 enabled TCP cork output buffering on all neo connections. In real-world use cases, where traffic is expected to be high, this is a sensible default. However, in _testing_ scenarios, where traffic is usually low, the 200ms delay in sending non-full packets is crippling.

For testing, we need to be able to disable output buffering -- all existing tests have been written assuming it is not present.

gavin-norman-sociomantic:
In the node, there is a [flag which activates TCP no-delay](https://github.com/sociomantic-tsunami/swarm/blob/00a171f48826c84a27093f1a2928b7b472cb3a3b/src/swarm/node/model/NeoNode.d#L603-L614). This should probably be adapted to disable TCP corking, instead, rather than mixing two different systems for output buffering.

PR #83, opened 2017-07-20
Allow for properly disabling socket output buffering

david-eckardt-sociomantic:
The recently added socket output buffering via `TCP_CORK` slows application tests down. So far a node can enable `TCP_NODELAY` by passing `no_delay = true` to the `Connection` constructor, but that is not enough: Both node and client need to enable `TCP_NODELAY` and disable `TCP_CORK`.
This PR adds a `bool no_delay` parameter to the client `Connection` constructor, extends its meaning to disable `TCP_CORK` in addition, and adds a `enableSocketNoDelay()` method to swarm neo clients.

To keep the code simpler `TCP_NODELAY` is now enabled after the authentication has completed, it used to be enabled from the beginning in the node. It should not make a difference because during handshake and authentication both client and node each wait for a response before sending the next message, which is precisely what Nagle's algorithm tries to accomplish.

### DIFF
--- a/src/swarm/neo/client/mixins/ClientCore.d
+++ b/src/swarm/neo/client/mixins/ClientCore.d
@@ -605,6 +605,22 @@ template ClientCore ( )
             static assert(false, "Invalid request type");
         }
     }
+
+    /***************************************************************************
+
+        Disables TCP socket output data buffering. Should be called before
+        adding a node.
+
+        Warning: This is meant to be used only in tests which perform sequential
+        requests. For a high data throughput rate it may impact performance so
+        do not use it in a production environment.
+
+    ***************************************************************************/
+
+    public void enableSocketNoDelay ( )
+    {
+        this.connections.socket_no_delay = true;
+    }
 }
 
 /*******************************************************************************
--- a/src/swarm/neo/connection/ConnectionBase.d
+++ b/src/swarm/neo/connection/ConnectionBase.d
@@ -263,7 +263,11 @@ abstract class ConnectionBase: ISelectClient
 
                 try
                 {
-                    this.outer.sender.cork = true;
+                    if (this.outer.no_delay)
+                        this.outer.socket.setsockoptVal(IPPROTO_TCP,
+                            socket.TcpOptions.TCP_NODELAY, true);
+                    else
+                        this.outer.sender.cork = true;
 
                     // Start the receive fiber, it will suspend itself immediately.
                     this.outer.recv_loop.start();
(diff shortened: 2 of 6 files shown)


nemanja-boric-sociomantic:
Should target `v3.3.x` instead.

gavin-norman-sociomantic:
This needs to go into v3.3.2, as the output buffering was enabled in v3.3.x.

src/swarm/neo/connection/ConnectionBase.d
```diff
@@ -683,13 +715,16 @@ abstract class ConnectionBase: ISelectClient
         Params:
             socket       = node/client connection socket
             epoll        = epoll select dispatcher for registering the socket
+            no_delay     = disables TCP socket output buffering
```
gavin-norman-sociomantic: Maybe add "(see comments on the `no_delay` field, above)"

src/swarm/neo/client/mixins/ClientCore.d
```diff
@@ -702,6 +702,22 @@ template ClientCore ( )
...
+        Disables TCP socket output data buffering. Should be called before
+        adding a node.
+
+        Warning: This is meant to be used only in tests which perform sequential
+        requests. For a high data throughput rate it may impact performance so
+        do not use it in a production environment.
+
+    ***************************************************************************/
+
+    public void enableSocketNoDelay ( )
```
gavin-norman-sociomantic, line 620: Though this PR will be included in a patch release, this change is a new feature. It should have release notes.

gavin-norman-sociomantic (changes requested):
I think no-delay should also be enabled for the client used in tests. (https://github.com/sociomantic-tsunami/swarm/blob/v4.x.x/test/neo/main.d#L57)

gavin-norman-sociomantic:
One of the commits should fix #80, too.

david-eckardt-sociomantic:
Rebased on v3.3.x.
> I think no-delay should also be enabled for the client used in tests.

That's right, but the tests exist only in v4.1 so it needs to be added when merging branches.

gavin-norman-sociomantic:
>That's right, but the tests exist only in v4.1 so it needs to be added when merging branches.

:+1: I'd do it as a separate PR, not as part of the merge commit, just to keep things simple to follow.

david-eckardt-sociomantic:
Updated, adding release note and extending the `no_delay` parameter documentation of the `ConnectionBase` constructor.

gavin-norman-sociomantic (approved):
Just needs "Fixes #80" in a commit message.

david-eckardt-sociomantic:
Updated the commit message to refer to #80.

gavin-norman-sociomantic:
Hm weird... conflicts...?

david-eckardt-sociomantic:
Conflict resolved.

david-eckardt-sociomantic:
For the record, there is frequent confusion about delayed sending of TCP packets/frames/segments through Nagle's algorithm and TCP Cork.
- Nagle's algorithm won't send a partial segment until the previous segment has been ACKnowledged.
    - Enabled by default, can be disabled via `TCP_NODELAY`.
- TCP Cork won't send a partial segment at all.
    - Disabled by default, can be enabled via `TCP_CORK`.
- TCP delayed ACK won't send a plain ACK segment. Instead it delays the ACK until there are data to send.
    - Enabled by default, can be disabled via `TCP_QUICKACK`, but the system may override and change it.

Of course a timeout applies to all. Some of them conflict with others:
- TCP Cork does the same as Nagle's algorithm (i.e. not sending partial segments) but is more strict, so enabling `TCP_CORK` defeats the effect of enabling `TCP_NODELAY`.
- TCP delayed ACK can escalate Nagle's algorithm if only Alice sends data to Bob, and Bob never sends data:
   - Alice uses Nagle's algorithm which waits for an ACK from Bob before sending a partial segment.
   - Bob uses delayed ACK which will send that ACK only together with data.
   - Bob never sends data so he never sends an ACK so Alice never receives an ACK so Alice never sends the partial segment.

PR #87, opened 2017-07-20
Update versioning scheme in README

### LINKED ISSUE #48: Reassess swarm versioning / maintenance policy
Swarm no longer contains concrete client classes and is thus no longer directly used by applications. It is only used indirectly via other libraries which implement concrete clients.

Thus, the set of versions of swarm that need to be maintained is defined solely by the needs of the dependent libraries. Keeping a strict, 6-month maintenance period for past major versions may not be necessary.

gavin-norman-sociomantic:
Fixes #48.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -112,9 +112,16 @@ for bug fixes that don't cause breaking changes.
 Support Guarantees
 ------------------
 
-* Major branch development period: 6 months
-* Maintained minor versions: 1 most recent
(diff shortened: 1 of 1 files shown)


gavin-norman-sociomantic:
Not 100% sure about this. What do people think?

mihails-strasuns-sociomantic (approved):

PR #90, opened 2017-07-26
Also retry all-nodes requests on I/O error

gavin-norman-sociomantic:
a005c1d3d2f3bb248af229502c5d8cebfa9ab219 inadvertently altered the
reconnection behaviour of all-nodes requests by only reconnecting after an
error event in epoll. We (obviously) also need to reconnect after a normal
connection I/O error (e.g. broken pipe).

### DIFF
--- a/src/swarm/neo/client/mixins/AllNodesRequestCore.d
+++ b/src/swarm/neo/client/mixins/AllNodesRequestCore.d
@@ -97,6 +97,7 @@ public struct AllNodesRequest ( Request, Connector, Disconnected, Initialiser,
 {
     import swarm.neo.client.RequestOnConn;
     import ocean.io.select.selector.EpollException;
+    import ocean.io.select.protocol.generic.ErrnoIOException;
 
     /// Policy object for handling connection establishment.
     private Connector connector;
@@ -140,8 +141,14 @@ public struct AllNodesRequest ( Request, Connector, Disconnected, Initialiser,
 
                 this.handler();
             }
-            // Only retry in the case of a connection error. Other errors
-            // indicate internal problems and should not be retried.
+            // Only retry in the case of a connection I/O error or error event
+            // in epoll. Other errors indicate internal problems and should not
+            // be retried.
+            catch ( IOError e )
+            {
+                this.disconnected(e);
+                reconnect = true;
+            }
             catch ( EpollException e )
             {
                 this.disconnected(e);


david-eckardt-sociomantic (approved):

src/swarm/neo/client/mixins/AllNodesRequestCore.d
```diff
@@ -140,8 +141,14 @@ public struct AllNodesRequest ( Request, Connector, Disconnected, Initialiser,
 
                 this.handler();
             }
-            // Only retry in the case of a connection error. Other errors
-            // indicate internal problems and should not be retried.
+            // Only retry in the case of a connection I/O error or error event
+            // in epoll. Other errors indicate internal problems and should not
+            // be retried.
+            catch ( IOError e )
```
nemanja-boric-sociomantic, line 147: Huh, I thought IOError was inherited from EpollException, at least that was my understanding at the time.

nemanja-boric-sociomantic: Here, so `IOError` inherits from `IOWarning` and `IOWarning` from `EpollException`: https://github.com/sociomantic-tsunami/ocean/blob/ba54399a79812c2be4c135a0597b88a93113f568/src/ocean/io/select/protocol/generic/ErrnoIOException.d#L75 https://github.com/sociomantic-tsunami/ocean/blob/ba54399a79812c2be4c135a0597b88a93113f568/src/ocean/io/select/protocol/generic/ErrnoIOException.d#L39

david-eckardt-sociomantic: `IOWarning` is a subclass of `ErnoException`.

gavin-norman-sociomantic: `ErrnoException`

nemanja-boric-sociomantic: :man_facepalming:

nemanja-boric-sociomantic: Still, there's a question about `IOWarning`. `IOWarning` should indicate remote party hangup. I think in this case we should reconnect as well.

gavin-norman-sociomantic: Hmm, the [description](https://github.com/sociomantic-tsunami/ocean/blob/ba54399a79812c2be4c135a0597b88a93113f568/src/ocean/io/select/protocol/generic/ErrnoIOException.d#L34-L35) of `IOWarning` doesn't sound like that, but I'm not sure where it's actually thrown, in practice.

PR #94, opened 2017-07-31
Call I/O wait to yield after 10 subsequent socket read() calls

david-eckardt-sociomantic:
Large numbers of subsequent non-blocking socket `read()` calls can succeed. The socket input fiber is then never suspended for a long time, blocking other pending activities. Correct this behaviour by suspending after `read()` has been called 10 times in a row without an I/O wait call in between.
This is a simple fix, we may come up with a better solution in the future.

### DIFF
--- a/src/swarm/neo/protocol/socket/MessageReceiver.d
+++ b/src/swarm/neo/protocol/socket/MessageReceiver.d
@@ -90,6 +90,23 @@ private class MessageReceiverBase
 
     protected size_t buffer_content_end = 0;
 
+    /***************************************************************************
+
+        The maximum number of subsequent `read` calls without an I/O-wait call.
+
+    ***************************************************************************/
+
+    private const uint read_iowait_max = 10;
+
+    /***************************************************************************
+
+        Counts the maximum number of subsequent `read` calls without an I/O-wait
+        call.
+
+    ***************************************************************************/
+
+    private uint read_iowait_count;
+
     /**************************************************************************/
 
     invariant ( )
@@ -402,10 +419,20 @@ private class MessageReceiverBase
             if (this.buffer.length < bytes_requested)
                 this.buffer.length = bytes_requested;
 
-            this.read();
+            auto event = Event.EPOLLIN;
+
+            if (++this.read_iowait_count >= this.read_iowait_max)
+            {
+                this.read_iowait_count = 0;
+                this.io_stats.num_iowait_calls++;
+                event = wait;
+            }
+
+            this.read(event);
 
             while (this.buffer_content_end < bytes_requested)
             {
+                this.read_iowait_count = 0;
                 this.io_stats.num_iowait_calls++;
                 this.read(wait);
             }


gavin-norman-sociomantic:
What about `write()`?

src/swarm/neo/protocol/socket/MessageReceiver.d
```diff
@@ -90,6 +90,23 @@ private class MessageReceiverBase
...
+    private const uint read_iowait_max = 10;
+
+    /***************************************************************************
+
+        Counts the maximum number of subsequent `read` calls without an I/O-wait
+        call.
+
+    ***************************************************************************/
+
+    private uint read_iowait_count;
```
gavin-norman-sociomantic, line 108: Does this need to be a class member? i.e. could it be a local function variable?

gavin-norman-sociomantic (approved):
Apart from the one question, LGTM.

david-eckardt-sociomantic:
> What about `write()`?

This will be a new feature in a separate PR. Currently protocol request handlers suspend their fiber during writing to accomplish the same behaviour, for convenience and consistency this will be moved into the neo I/O engine in `swarm.neo`. For reading such a solution outside the `swarm.neo` I/O engine is not possible so it has to be a bugfix.

nemanja-boric-sociomantic (approved):

src/swarm/neo/protocol/socket/MessageReceiver.d
david-eckardt-sociomantic: It could in principle be a local function variable, but it's not so easy. The loop which eventually calls `read()` repeatedly is in [`ConnectionBase.ReceiveLoop.fiberMethod`](https://github.com/sociomantic-tsunami/swarm/blob/v3.3.x/src/swarm/neo/connection/ConnectionBase.d#L560). I see two ways:
- Either the logic of suspending after `receiver.receive` has been called 10 times without calling the `cast(Event)this.suspend(...)` callback would have to go in this loop, relying on internal behaviour of `MessageReceiver`,
- or a counter method parameter would need to be added to `MessageReceiver.receive`, all the way down to where `read` is actually called.

Both ways look like a hassle to me.

gavin-norman-sociomantic: Hmm, I don't get it. Currently, `read_iowait_count` is only accessed/modified inside one method: `ensureMinimumAmountOfBytesInBuffer`.

david-eckardt-sociomantic: Yes, but a local counter variable must be stored in the loop. The loop is not in `ensureMinimumAmountOfBytesInBuffer` but in `ConnectionBase.ReceiveLoop.fiberMethod`, the indirect caller. A local counter in `ensureMinimumAmountOfBytesInBuffer` would be 0 on each repeated call.

gavin-norman-sociomantic: Oh ok yeah I see. Fine then.

PR #104, opened 2017-08-09
Add neo client config struct & ctor

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Will add a test case reading the `Config` instance from a file.

gavin-norman-sociomantic:
Updated with test case.

scott-gibson-sociomantic:
I can't speak for other team's apps but i know that reporting apps take advantage of the previous class based IClient.Config by inheriting and adding things such as throttle information, channels to read etc. 
Do you want to limit that behaviour here and prefer apps to separate the config?

gavin-norman-sociomantic:
Interesting. I went for a struct as I couldn't imagine why anyone would need to inherit the config container. Tbh it sounds to me like what you describe is... kind of weird...? I mean, throttling is a separate issue to the basic configuration of a swarm client.

That said, my intention wasn't to deliberately limit usage, so maybe I should change it to a class.

gavin-norman-sociomantic:
Updated. Config is now a class.

gavin-norman-sociomantic:
>Interesting. I went for a struct as I couldn't imagine why anyone would need to inherit the config container.

And we consciously avoid classes and inheritance, generally, in neo.

scott-gibson-sociomantic:
`Interesting. I went for a struct as I couldn't imagine why anyone would need to inherit the config container.`

Legacy does this itself with the Scheduling*Client config. I'm not sure how neo approaches this sort of behaviour but it's worth a thought.

scott-gibson-sociomantic:
For example in TubeEye,  i have my dmq input class taking a config instance of
```
public class Dmq : SchedulingDmqClient.Config
{
    /// The Dmq channel to output processed records to.
    public ConfigFiller.Required!(char[]) channel;
}
```

gavin-norman-sociomantic:
>Legacy does this itself with the Scheduling*Client config. I'm not sure how neo approaches this sort of behaviour but it's worth a thought.

Right, this was necessary internally in swarm with the legacy client. This is not necessary with the neo client and is indeed shunned, as I mentioned above. That's why I thought it was unnecessary.

scott-gibson-sociomantic:
Again, i would like some input from the other teams on if they do a similar thing. It's hard to search github :(

gavin-norman-sociomantic:
Yep. It does no harm being a class, though, so whether a certain usage is categorised by me as "weird" or not doesn't really matter ;)

src/swarm/neo/client/mixins/ClientCore.d
```diff
@@ -92,6 +92,24 @@ template ClientCore ( )
 
     /***************************************************************************
 
+        Config struct which may be passed to the ctor. Designed for use with
+        ocean's ConfigFiller helper.
+
+    ***************************************************************************/
+
+    public static class Config
```
nemanja-boric-sociomantic, line 100: commit message is outdated, this is a `class`.

src/swarm/neo/client/mixins/ClientCore.d
```diff
@@ -92,6 +92,24 @@ template ClientCore ( )
 
     /***************************************************************************
 
+        Config struct which may be passed to the ctor. Designed for use with
```
nemanja-boric-sociomantic: `struct`

relnotes/config.feature.md
```diff
@@ -0,0 +1,6 @@
+* `swarm.neo.client.mixins.ClientCore`
+
+  A new constructor has been added that accepts a ``Config`` struct instance
```
nemanja-boric-sociomantic: `s/struct/class` in multiple places I guess.

nemanja-boric-sociomantic:
Generally, LGTM

relnotes/config.feature.md
gavin-norman-sociomantic: Updated.

gavin-norman-sociomantic: Oh wait, I missed the commit message.

gavin-norman-sociomantic: Updated.

nemanja-boric-sociomantic: This is still `Config` struct instance.

gavin-norman-sociomantic: Ha I missed this one completely ;)

gavin-norman-sociomantic: Updated again.

PR #107, opened 2017-08-09
Correct module header references to license file

gavin-norman-sociomantic:
It's called LICENSE.txt, not LICENSE_BOOST.txt.


nemanja-boric-sociomantic (approved):
Yes, looks like this is copy/pasted from Ocean, where we have more licenses.

gavin-norman-sociomantic:
Right.

PR #111, opened 2017-08-11
Log stats on neo connections

gavin-norman-sociomantic:


src/swarm/node/model/INodeInfo.d
```diff
@@ -78,6 +78,40 @@ public interface INodeInfo
...
+
+    public size_t num_neo_connections ( );
+
+
+    /**************************************************************************
+
+        Returns:
+            the number of active neo connections currently being handled
+            (Note: the number of idle connections in the pool can be calculated
+            by num_connections - num_open_connections)
```
nemanja-boric-sociomantic: `by num_neo_connections - num_open_neo_connections`

src/swarm/node/model/Node.d
```diff
@@ -241,6 +241,21 @@ public abstract class INodeBase : INode, INodeInfo
...
+    /**************************************************************************
+
+        Returns:
+            the limit of the number of neo connections (i.e. the maximum number
+            of connections the node can handle in parallel) or 0 if limitation
+            is disabled
+
+     **************************************************************************/
+
+    override public size_t neo_connection_limit ( )
```
nemanja-boric-sociomantic, line 253: Bah, this is useless, could we do something about this? Maybe make the `INodeInfo` abstract class and provide default implementation? I'm happy to leave it as is and wait for better days for the swarm's node framework.

gavin-norman-sociomantic: Puh... yeah... I wanted to mess with the horrible inheritance hierarchy as little as possible in this PR.

src/swarm/node/model/INodeInfo.d
gavin-norman-sociomantic: Updated.

nemanja-boric-sociomantic (approved):

PR #112, opened 2017-08-16
Add new combined neo/legacy node and connection bases

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Work-in-progress on a set of simplified base classes for nodes and connection handlers. The principle changes are:
1. Avoiding the use of informational interfaces to hide implementation details. In the node, where nothing is exposed to "user code", we don't care about this.
2. Melding neo functionality into the base node.

gavin-norman-sociomantic:
Still needs tidying up, testing (including with *protos and logging utils), and release notes.

gavin-norman-sociomantic:
@nemanja-boric-sociomantic if you fancy having a look at this, in general terms, feel free to make suggestions.

nemanja-boric-sociomantic:
I will take a look :+1:

src/swarm/node/simplified/NodeTemplate.d
```diff
@@ -0,0 +1,220 @@
...
+
+    Node base template.
+
+    Template params:
+        ConnHandler = type of connection handler (the node contains a
+            SelectListener instance which owns a pool of instances of this type)
+
+*******************************************************************************/
+
+public class NodeTemplate ( ConnHandler : LegacyConnectionHandlerBase ) : NodeBase
```
nemanja-boric-sociomantic, line 37: I wonder if this one or `NodeBase` are necessary? Wouldn't it be simpler if we just meld them together. Would we ever use `NodeBase` on its own?

gavin-norman-sociomantic: This was the first thing I wondered. `NodeBase` is used, though: it's passed to the connection handlers, in this new framework. The other approach would be to make the connection handler templated on the type of node (i.e. an instantiation of `NodeTemplate`). I _think_ this would end up more complicated.

nemanja-boric-sociomantic: Ah, yes, I see. Yeah, that makes sense.

gavin-norman-sociomantic: I'll document this reasoning.

gavin-norman-sociomantic:
Updated:
* Improved documentation.
* Refactored a bunch more stuff into `NodeBase`, out of `NodeTemplate`.
* Test node no longer needs to support the legacy protocol.

gavin-norman-sociomantic:
Updated:
* Integrated stats logging.

gavin-norman-sociomantic:
Green now :)

I'm just trying to adapt dhtnode to use this, to check it works as expected.

gavin-norman-sociomantic:
>I'm just trying to adapt dhtnode to use this, to check it works as expected.

:( Fun with legacy shared resources :(

gavin-norman-sociomantic:
Blocked again. I tried adapting the dhtnode to use this and had severe problems with the legacy shared resources. It may be that this clean-up isn't really feasible until we go neo-only, unfortunately.

gavin-norman-sociomantic:
Sadly, despite the work I put into this, I think it's best to close this PR. It doesn't seem feasible to address this issue until we remove support for the legacy protocol.

PR #113, opened 2017-08-16
Don't pass empty channel name sentinel to the user for getChannels

nemanja-boric-sociomantic:
In legacy GetChannels request, every node will send the list of the
channels, followed by the empty entry, which serves as a sentinel. This
shouldn't be passed to the user.

### DIFF
--- a/src/swarm/client/request/GetChannelsRequest.d
+++ b/src/swarm/client/request/GetChannelsRequest.d
@@ -82,15 +82,16 @@ public scope class GetChannelsRequestTemplate ( Base : IRequest, Resources,
     {
         auto output = this.params.io_item.get_node_value();
 
-        do
-        {
-            this.reader.readArray(*this.resources.channel_buffer);
+        this.reader.readArray(*this.resources.channel_buffer);
 
+        while ( this.resources.channel_buffer.length )
+        {
             output(this.params.context, this.resources.conn_pool_info.address,
                 this.resources.conn_pool_info.port,
                 *this.resources.channel_buffer);
+
+            this.reader.readArray(*this.resources.channel_buffer);
         }
-        while ( this.resources.channel_buffer.length );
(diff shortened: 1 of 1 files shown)


src/swarm/client/request/GetChannelsRequest.d
```diff
@@ -82,15 +82,16 @@ public scope class GetChannelsRequestTemplate ( Base : IRequest, Resources,
     {
         auto output = this.params.io_item.get_node_value();
 
-        do
-        {
-            this.reader.readArray(*this.resources.channel_buffer);
+        this.reader.readArray(*this.resources.channel_buffer);
 
+        while ( this.resources.channel_buffer.length );
```
mathias-lang-sociomantic: Luckily DMD errors on this.

nemanja-boric-sociomantic: DMD has this nasty property to error when it shouldn't and not error when it should ;-)

nemanja-boric-sociomantic:
Updated.

gavin-norman-sociomantic:
Wow, I'd completely forgotten that we even have any request implementation code in swarm! :laughing: 

What's the impetus for changing this now? Our general policy is not to change anything in the legacy code unless it's an absolute emergency and there's no work-around.

nemanja-boric-sociomantic:
I think the solution is easy to workaround - just filter the empty channel in the notifier, but it's just one if these things people been reporting every few months and it's trivial to do, and I don't think fixing it could cause any issues. I think the closing this PR would be appropriate as well as merging it, but I slightly prefer closing it now you mentioned this.

gavin-norman-sociomantic:
Yeah, it's just that any small behavioural change like this could introduce new bugs, leading us into a maintenance cycle. Not worth it.

PR #115, opened 2017-08-24
Fixes for bugs resuming request fibers when connection send loop exits

gavin-norman-sociomantic:
I'd suggest we hold off merging this until @david-eckardt-sociomantic has a look.

### DIFF
--- a/src/swarm/neo/connection/ConnectionBase.d
+++ b/src/swarm/neo/connection/ConnectionBase.d
@@ -77,6 +77,11 @@ abstract class ConnectionBase: ISelectClient
 
     private class SendLoop : MessageFiber
     {
+        /// Exception which caused the send loop to exit. A reference to it is
+        /// stored in order to rethrow in the context of the caller of
+        /// registerForSending().
+        private Exception send_loop_exited_ex;
+
         /***********************************************************************
 
             Token used when suspending/resuming the fiber.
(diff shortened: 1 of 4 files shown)


src/swarm/neo/node/Connection.d
```diff
@@ -256,8 +256,6 @@ class Connection: ConnectionBase
     override protected void notifyShutdown ( RequestId id, Exception e )
```
nemanja-boric-sociomantic, line 256: re: commit message - perhaps it's worth saying why there's there a running request fiber while the send loop's fiber is also running?

gavin-norman-sociomantic: Is this not enough?
>This will be the request which resumed the connection's send loop by calling SendLoop.registerForSending.

nemanja-boric-sociomantic: Yeah, this should be enough :+1:

src/swarm/neo/connection/ConnectionBase.d
```diff
@@ -156,6 +189,11 @@ abstract class ConnectionBase: ISelectClient
                 if ( this.idle )
                 {
                     this.resume(this.fiber_token.get(), this.outer);
+
+                    // If the send loop exited, notify the calling fiber. (See
+                    // note on error handling, above.)
+                    if ( !this.loop_started )
+                        throw send_loop_exited_ex;
```
david-eckardt-sociomantic: Hm, wouldn't it make more sense to throw the exception which cause the send loop to exit?

gavin-norman-sociomantic: Yeah... We'd have to dup it I guess...? By the time we get to this point, the exception no longer exists.

gavin-norman-sociomantic: Updated. (Retested with https://github.com/sociomantic-tsunami/dhtnode/issues/13.)

david-eckardt-sociomantic: For the record, the code has already been updated: Making `send_loop_exited_ex` non-static, assigning the exception caught in `fiberMethod` and rethrowing it here is sufficient. The exception doesn't need to be dupped.

david-eckardt-sociomantic: Two more things:
- `throw this.send_loop_exited_ex;`
- Reset `this.send_loop_exited_ex = null` after throwing (e.g. in a `scope (exit)` clause).

gavin-norman-sociomantic: Updated. (Also added an assert that the exception is non-null before throwing it.)

david-eckardt-sociomantic (approved):

gavin-norman-sociomantic:
Anyone feel like merging?

PR #117, opened 2017-08-24
Do not resume yielded RoCs that are removed from the list while iterating

gavin-norman-sociomantic:
(Based on #115.)

### DIFF
--- a/src/swarm/neo/connection/YieldedRequestOnConns.d
+++ b/src/swarm/neo/connection/YieldedRequestOnConns.d
@@ -20,6 +20,7 @@ module swarm.neo.connection.YieldedRequestOnConns;
 
 *******************************************************************************/
 
+import ocean.transition;
 import ocean.io.select.client.SelectEvent;
 
 /******************************************************************************/
@@ -143,6 +144,8 @@ class YieldedRequestOnConns: ISelectEvent
 
     private static struct YieldedQueue
     {
+        import ocean.core.Array : contains;
+
         /***********************************************************************
 
             The two queues of yielded `RequestOnConn`s.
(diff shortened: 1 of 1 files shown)


gavin-norman-sociomantic:
Rebased.

david-eckardt-sociomantic:
I'm surprised that this patch is necessary. The `TreeQueue` should support removing during iteration IIRC -- looks like a bug in the `TreeQueue` to me. I'll take a closer look at the `TreeQueue` to see if my assumption is correct.

david-eckardt-sociomantic:
Referring to https://github.com/sociomantic-tsunami/dhtnode/issues/14#issuecomment-324643237:

> It seems to me that it must be possible to (at least) *remove* items from the queue being iterated over. (This would fix the current problem, presumably.)

That should indeed have the same effect as the `removed_while_iterating` list because the `TreeQueue`s of yielded RoCs support removing during iteration. Could you please try that? It should be as simple as

```d
public bool remove ( IYieldedRequestOnConn roc )
{
        this.queue[!this.active].remove(roc);
        return this.queue[this.active].remove(roc);
}
```

> I also wonder, in general whether the careful swapping of queues is really necessary...

Now I wonder, too, if they need to be removed from both queues. But if I remember (my own design :blush:) correctly, it is about pushing RoCs during iteration: If a request yields again after it has been resumed then `swapAndPop` may resume it again within the same call i.e. without an intermediate event loop cycle. To be honest I'm not entirely sure if this is in fact possible -- thinking about it I think it shouldn't, but it requires a unit test to be sure.

gavin-norman-sociomantic:
Replaced with #121.

PR #118, opened 2017-08-25
Make IoVecTracker GC friendly

nemanja-boric-sociomantic:
Previously, array of IoVecTracker were reset to `null`, making it allocate every time it's constructed. Also, some `enableStompings` are put in places where the length is shrinking.


gavin-norman-sociomantic:
Oof nasty. Good catch!

gavin-norman-sociomantic (approved):

PR #123, opened 2017-08-31
Add warning about array casting

gavin-norman-sociomantic:


src/swarm/neo/util/AcquiredResources.d
```diff
@@ -66,6 +66,26 @@ public struct AcquiredArraysOf ( T )
...
+            struct S { int i; hash_t h; }
+            auto s_array = cast(S[]*)void_array;
+            s_array.length = 23;
+        ---
+
+        The exact reason for this is not known, but it appears to lead to
+        corruption of internal GC data (possibly type metadata associated with
+        the array's pointer). Setting the length of the array *before* casting
+        is safe.
+
```
david-eckardt-sociomantic: Suggest adding a positive example, such as
```d
(*void_array).length = 23 * S.sizeof; // First set the buffer length...
auto s_array = cast(S[])*void_array;   // ... then cast the array.
```
Maybe even change the order: First say how one should do it, then mention a way that is known to fail.

gavin-norman-sociomantic: Yep, good idea.

gavin-norman-sociomantic: Updated.

gavin-norman-sociomantic:
Rebased.

david-eckardt-sociomantic (approved):

PR #131, opened 2017-09-08
Safe usage of void[]s cast to other array types in neo.util.AcquiredResources.d

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Unfortunately the bug fix requires a significant new feature :/

gavin-norman-sociomantic:
First commit is still WIP, pending comment.

gavin-norman-sociomantic:
One option, as we've not encountered this bug in any existing code, would be to move this to a minor release. I dunno... it is a bug fix, but involves new features and quite some refactoring.

src/swarm/neo/util/VoidBufferAsArrayOf.d
```diff
@@ -0,0 +1,159 @@
...
+    public int opApply ( int delegate ( ref T ) dg )
+    {
+        int res;
+        for ( size_t i = 0; i < this.length; i++ )
+        {
+            res = dg(*this.opIndex(i));
+            if ( res )
+                break;
+        }
+        return res;
```
david-eckardt-sociomantic: Suggest
```d
foreach (ref element; cast(T[])*this.buffer)
    if (int stop = dg(element))
        return stop;
return 0;
```

gavin-norman-sociomantic: _That's not the opApply recipe!!!_

(Yeah.)

src/swarm/neo/util/VoidBufferAsArrayOf.d
```diff
@@ -0,0 +1,159 @@
...
+    }
+
+    public void opIndexAssign ( T elem, size_t t_index )
+    {
+        *this.opIndex(t_index) = elem;
+    }
+
+    public T[] opSlice ( )
+    {
+        return (cast(T*)this.buffer.ptr)[0..this.length];
```
david-eckardt-sociomantic: `cast(T[])(*this.buffer)`?

david-eckardt-sociomantic:
There is a way of making `VoidBufferAsArrayOf` simplistic: Given there is `opSlice()` many operators can be derived by the user from it:
```d
VoidBufferAsArrayOf!(int) array;
int x = array[][3]; // instead of array[3], saves opIndex
int* x_ptr = &array[][3]; // instead of &array[3]
array[][2] = 4711; // instead of array[2] = 4711, saves opIndexAssign
int[] subarray = array[][3 .. 7]; // instead of array[3 .. 7], saves opSlice overload
foreach (i, element; array[]) {} // saves four opApply+Reverse methods
```
This may look a bit confusing, but instead of `opIndexAssign()` you call it `array()`, for example.

david-eckardt-sociomantic:
> There is a way of making `VoidBufferAsArrayOf` simplistic...

My point is, special logic is required only to change the length of the buffer. Everything else the user can do with a `cast(T[])(*buffer)`.

nemanja-boric-sociomantic:
I feel this will just move complexity to the user.

gavin-norman-sociomantic:
Right, makes sense.

nemanja-boric-sociomantic:
I'm saying this because the code is templated, and always slicing it will save you constant amount of the implementation code, whereas user would have to deal with the "do I need to slice or not in this context?" every time it's using it, instead of just dealing with it as a normal array (modulo `.length`).

david-eckardt-sociomantic:
> I feel this will just move complexity to the user.

Yes, the user needs to prefix all operations with `array` or `getArray()`. But the advantage is that the user can do what s/he wants rather than requiring support for every single array operation. For some operations like `&array[i]` cannot even be overloaded (and I wouldn't use `opIndex` for that). With explicit overloading we'd go down the road of adding a method for every single operation someone has ever used.

nemanja-boric-sociomantic:
> But the advantage is that the user can do what s/he wants rather than requiring support for every single array operation

I agree.

gavin-norman-sociomantic:
Updated with simplified wrapper struct.

src/swarm/neo/util/VoidBufferAsArrayOf.d
```diff
@@ -0,0 +1,93 @@
...
+        this.buffer.length = len * T.sizeof;
+        enableStomping(*this.buffer);
+    }
+
+    public void opCatAssign ( T elem )
+    {
+        this.length = this.length + 1;
+        auto elem_in_array =
+            cast(T*)this.buffer.ptr + ((this.length - 1) * T.sizeof);
+        *elem_in_array = elem;
```
david-eckardt-sociomantic: You can just append to `(*this.buffer)`, and there should also be an `opCatAssign` overload to append an array:
```d
T[] opCatAssign ( in T[] arr )
{
    return cast(T[])((*this.buffer) ~= arr);
}
```
`arr` is automatically cast to `void[]`. (The real `~=` operator returns the array.) With this overload appending a single element is as simple as
```d
T[] opCatAssign ( in T element )
{
    return this.opCatAssign((&element)[0 .. 1]);
}
```

src/swarm/neo/util/AcquiredResources.d
```diff
@@ -88,15 +100,17 @@ public struct AcquiredArraysOf ( T )
...
+            this.acquired = VoidBufferAsArrayOf!(void[])(&this.buffer);
         }
 
         // Acquire and re-initialise new buffer to return to the user. Store
         // it in the container buffer.
-        this.acquired ~= cast(T[])newBuffer(4);
-        return &this.acquired[$-1];
+        this.acquired ~= newBuffer(T.sizeof * 4);
+
+        return cast(T[]*)this.acquired.array()[this.acquired.length-1];
```
david-eckardt-sociomantic:
```d
void[][] array_of_void_slices = this.acquired.array();
auto array_of_T_slices = cast(T[][])array_of_void_slices;
return &array_of_T_slices[$ - 1];
```
General advice to avoid corruption trouble: Don't cast pointers to arrays. Instead cast the array first, then use the `&` operator.

src/swarm/neo/util/AcquiredResources.d
```diff
@@ -277,15 +300,16 @@ public struct Acquired ( T )
...
-            this.acquired = cast(Elem[])newBuffer(Elem.sizeof * 4);
+            this.buffer = newBuffer(Elem.sizeof * 4);
+            this.acquired = VoidBufferAsArrayOf!(Elem)(&this.buffer);
         }
 
         // Acquire new element.
         this.acquired ~= this.t_pool.get(new_t);
 
-        return this.acquired[$-1];
+        return this.acquired.array()[this.acquired.length-1];
```
david-eckardt-sociomantic: `[$ - 1]`

gavin-norman-sociomantic:
Updated.

gavin-norman-sociomantic:
Updated:
* Changed `opCatAssign` as @david-eckardt-sociomantic suggested.
* Fully documented.
* Added unittest and usage example.

src/swarm/neo/util/VoidBufferAsArrayOf.d
```diff
@@ -0,0 +1,210 @@
...
+            arr = elements to append
+
+        Returns:
+            a slice of this array, now with the specified elements appended
+
+    ***************************************************************************/
+
+    public T[] opCatAssign ( in T[] arr )
+    {
+        return cast(T[])((*this.buffer) ~= arr);
```
david-eckardt-sociomantic: This needs `cast(void[])arr` i.e. casting `const` away from `arr` and its elements.

To avoid undefined behaviour (in particular, if `T` contains `const` references, modifying the referenced objects) the elements of `(*this.buffer)` may only be accessed through `cast(T[])(*this.buffer)`, which restores the correct `const`ness of `T`. Making sure of that is the whole point of this struct, but maybe state this in the documentation of the `buffer` field.

gavin-norman-sociomantic: The constness of `T` is a different question to the consteness of this method's argument, or?

david-eckardt-sociomantic: Yes, but the constness of `T` is what matters most. We only read `arr` to store a copy of it, but `in` makes the elements of `arr` always `const`, no matter if `T` is `const` or not.

gavin-norman-sociomantic: Updated.

src/swarm/neo/util/VoidBufferAsArrayOf.d
```diff
@@ -0,0 +1,215 @@
...
+        this.buffer.length = len * T.sizeof;
+        enableStomping(*this.buffer);
+    }
+
+    /***************************************************************************
+
+        Appends an array of elements.
+
+        Note that mutable copies of appended elements are made internally but
+        are inaccessible from the outside.
```
david-eckardt-sociomantic: Better, "... but to access them from the outside the constness of T applies."

gavin-norman-sociomantic: Updated.

david-eckardt-sociomantic (approved):

gavin-norman-sociomantic:
What do we think about the question of whether to do this in a patch vs feature release? It _is_ a bug fix, but it's not something that's come up, in practice.

gavin-norman-sociomantic:
Moving to a feature branch.

gavin-norman-sociomantic:
Rebased on v4.x.x.

gavin-norman-sociomantic:
Recreated as https://github.com/sociomantic-tsunami/swarm/pull/136.

PR #132, opened 2017-09-10
Infra upgrade for v5.x.x

mihails-strasuns-sociomantic:
- Upgrade to HEAD of ocean v4.x.x
- Start testing with dmd-transitional 2.071.2.s2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,8 @@ env:
         - DMD=1.080.0 DIST=xenial F=devel
         - DMD=2.070.2.s12 DIST=xenial F=production
         - DMD=2.070.2.s12 DIST=xenial F=devel
+        - DMD=2.071.2.s2 DIST=xenial F=production
+        - DMD=2.071.2.s2 DIST=xenial F=devel
 
 install: beaver dlang install
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,12 +19,12 @@ env:
         # Make sure beaver is in the PATH
         - PATH="$(git config -f .gitmodules submodule.beaver.path)/bin:$PATH"
     matrix:
-        - DMD=1.080.0 DIST=xenial F=production AFTER_SCRIPT=1
-        - DMD=1.080.0 DIST=xenial F=devel
-        - DMD=2.070.2.s12 DIST=xenial F=production
-        - DMD=2.070.2.s12 DIST=xenial F=devel
-        - DMD=2.071.2.s2 DIST=xenial F=production
-        - DMD=2.071.2.s2 DIST=xenial F=devel
+        - DMD=1.081.* DIST=xenial F=production AFTER_SCRIPT=1
+        - DMD=1.081.* DIST=xenial F=devel
+        - DMD=2.070.2.s* DIST=xenial F=production
+        - DMD=2.070.2.s* DIST=xenial F=devel
+        - DMD=2.071.2.s* DIST=xenial F=production
+        - DMD=2.071.2.s* DIST=xenial F=devel
 
 install: beaver dlang install
 
--- a/src/swarm/node/model/NeoNode.d
+++ b/src/swarm/node/model/NeoNode.d
@@ -53,7 +53,7 @@ import ocean.io.compress.lzo.LzoChunkCompressor;
 
 import ocean.util.container.pool.model.IAggregatePool;
 
-import ocean.util.log.Log;
+import ocean.util.log.Logger;
 
 import ocean.core.Enforce;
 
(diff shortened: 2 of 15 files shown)


mihails-strasuns-sociomantic:
Anyone? :(

gavin-norman-sociomantic:
Could you put the subsum output in the commit message?

gavin-norman-sociomantic:
Also, does this imply an update to the dependencies listed in the README?

david-eckardt-sociomantic (approved):

src/swarm/node/model/NeoNode.d
```diff
@@ -870,7 +870,8 @@ public class NodeBase ( ConnHandler : ISwarmConnectionHandler ) : INodeBase
     ***************************************************************************/
 
     private void handleUpdateCredentials ( cstring args,
-        void delegate ( cstring response ) send_response )
+        void delegate ( cstring response ) send_response,
+        void delegate ( ref mstring response ) wait_reply )
```
gavin-norman-sociomantic, line 874: @nemanja-boric-sociomantic do we need to change anything more than this here?

test/neo/client/Client.d
```diff
@@ -31,7 +31,7 @@ public class Client
 
     ***************************************************************************/
 
-    private class Neo
+    public class Neo
```
gavin-norman-sociomantic, line 34: Huh. This is not intended to be public. Why is this change required?

gavin-norman-sociomantic (commented):
Hmm... the protection attribute changes are interesting...

.travis.yml
```diff
@@ -23,6 +23,8 @@ env:
         - DMD=1.080.0 DIST=xenial F=devel
         - DMD=2.070.2.s12 DIST=xenial F=production
         - DMD=2.070.2.s12 DIST=xenial F=devel
+        - DMD=2.071.2.s2 DIST=xenial F=production
+        - DMD=2.071.2.s2 DIST=xenial F=devel
```
nemanja-boric-sociomantic: Can we use wildcard here?

src/swarm/node/model/NeoNode.d
nemanja-boric-sociomantic: No, I think that's it. It just enables you to call `wait_reply` from the handler delegate. No more work is needed.

test/neo/client/Client.d
mihails-strasuns-sociomantic: It is used from `test/neo/main.d`

mihails-strasuns-sociomantic: also `test/client_auth/main.d`

mihails-strasuns-sociomantic:
> Also, does this imply an update to the dependencies listed in the README?

It depends on HEAD branch state so I am unsure what to put into README. Makes more sense to update it when cleaning up to use actual release eventually.

test/neo/client/Client.d
gavin-norman-sociomantic: Ah yes, of course. It's used for the various aliases it contains. I think this should be a case for a private ctor.

gavin-norman-sociomantic:
>Makes more sense to update it when cleaning up to use actual release eventually.

:+1:

test/neo/client/Client.d
gavin-norman-sociomantic: `private class Blocking` will need the same treatment.

gavin-norman-sociomantic: Actually, no, sorry, it only contains methods, which can be accessed via the instance.

gavin-norman-sociomantic: Actually, no, sorry 2: `Blocking` mixes is `TaskBlockingCore`, which contains types which will need to be accessible from outside.

mihails-strasuns-sociomantic: Whatever. Can you do any change you find necessary as a separate PR? I am only fixing compiler errors for existing code :P

mihails-strasuns-sociomantic:
Updated matrix with wildcard

test/neo/client/Client.d
gavin-norman-sociomantic: Sure. Can you at least change the ctor of `Neo` to be private in the same commit where you make the class public?

mihails-strasuns-sociomantic: It is already private because it is mixed in by `ClientCore`

gavin-norman-sociomantic: Ah excellent. Thanks for checking.

gavin-norman-sociomantic (approved):

PR #138, opened 2017-09-19
Add new AcquiredArraysOf method to return safely wrapped arrays

gavin-norman-sociomantic:


gavin-norman-sociomantic:
About the deprecation path: it's annoying that we end up with different method names (`acquire` vs `acquireWrapped`) for different `T`. I'd envisage deprecating and renaming `acquireWrapped` to `acquire`, in the next major.

david-eckardt-sociomantic (approved):

gavin-norman-sociomantic:
Auto-merged after David's approval.

PR #140, opened 2017-09-19
Switch to formatter/logger

mihails-strasuns-sociomantic:


gavin-norman-sociomantic:
LGTM but Travis says no.

mihails-strasuns-sociomantic:
Green

src/swarm/neo/protocol/ProtocolError.d
```diff
@@ -90,11 +93,18 @@ class ProtocolError: Exception
...
 
-    private void setFmt_ ( istring file, long line, cstring fmt, ... )
+    private void setFmt_ ( T... ) ( istring file, long line, cstring fmt, T args )
     {
-        this.reused_msg.length = 0;
-        auto msg = this.reused_msg[];
-        Format.vformat(msg, fmt, _arguments, _argptr);
+        this.reused_msg.reset();
+        sformat(
+            cast(FormatterSink) (cstring chunk) {
```
mathias-lang-sociomantic, line 100: Nasty. A `static if` would have been better than a `cast` IMO.

mihails-strasuns-sociomantic: It will all be gone in v5.x.x anyway (already did the same changes there). Casting to modify return type is a proven technique :P

mathias-lang-sociomantic: Not anymore, dmd errors on that code nowadays :grumpy:

PR #146, opened 2017-10-11
Replace deprecate import

daniel-zullo:
Import core.thread instead of ocean.core.Thread.


mathias-lang-sociomantic:
That was done in v4.x.x already

daniel-zullo:
:S closing it then

PR #149, opened 2017-10-13
Fix `RequestEventDispatcher.eventLoop` not exiting

david-eckardt-sociomantic:
`dispatchQueuedSignals` can continue waiting fibers so the event loop could call `nextEvent` with no waiting fiber. Instead, if there is no waiting fiber left, the event loop should exit.

### DIFF
--- a/src/swarm/neo/request/RequestEventDispatcher.d
+++ b/src/swarm/neo/request/RequestEventDispatcher.d
@@ -533,6 +533,8 @@ public struct RequestEventDispatcher
 
             if ( this.waitingWriters() )
                 writer = this.popWaitingWriter();
+            else if ( !this.waiting_fibers.length )
+                break;
 
             this.handleNextEvent(conn, writer);
         }


src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -533,6 +533,8 @@ public struct RequestEventDispatcher
 
             if ( this.waitingWriters() )
                 writer = this.popWaitingWriter();
+            else if ( !this.waiting_fibers.length )
```
gavin-norman-sociomantic, line 536: I do think the same fix is required in `handleNextEvent` (line 590).

david-eckardt-sociomantic: It is not required:
- If there is no waiting fiber then `eventLoop` won't call `handleNextEvent`.
- On the first `handleNextEvent` loop cycle `dispatchQueuedSignals` won't do anything because it has just been called in `eventLoop`.
- The loop does the next cycle only if there is a writer so even if the next `dispatchQueuedSignals` call in the loop removes all other waiting fibers we know we have a waiting writer.

gavin-norman-sociomantic (approved):

PR #150, opened 2017-10-13
Use constant id for debug output in legacy RequestConnection

### LINKED ISSUE #148: Use constant identifiers for debug printouts of RequestConnections
`RequestConnection` right now uses `object_pool_index` for the `SwarmDebug` printouts. This is problematic, since these numbers are not constant and can change during the request handling (when recycling _other_ `RequestConnection` objects), making it hard to debug. It would be easier if the unique numbers were assigned for the `RequestConnection` objects.

https://github.com/sociomantic-tsunami/swarm/blob/v4.x.x/src/swarm/client/connection/RequestConnection.d#L454-L455

gavin-norman-sociomantic:
The object_pool_index that is currently used can change, when items are
added or removed from the object pool. This is very confusing during
debugging!

Fixes #148.

### DIFF
--- a/src/swarm/client/connection/RequestConnection.d
+++ b/src/swarm/client/connection/RequestConnection.d
@@ -110,6 +110,24 @@ public abstract class IRequestConnection :
     public size_t object_pool_index;
 
 
+    /***************************************************************************
+
+        Uniquely identifying integer for this instance.
+
+    ***************************************************************************/
+
+    private Immut!(size_t) id;
+
+
+    /***************************************************************************
+
+        Count of instances of this class.
+
+    ***************************************************************************/
+
+    static private size_t instance_count;
+
+
     /**************************************************************************
 
         This alias for chainable methods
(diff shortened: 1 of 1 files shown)


src/swarm/client/connection/RequestConnection.d
```diff
@@ -364,6 +382,8 @@ public abstract class IRequestConnection :
         assert(conn_pool !is null, typeof(this).stringof ~ ".ctor: connection pool interface must be non-null");
         assert(params !is null, typeof(this).stringof ~ ".ctor: request params instance must be non-null");
 
+        this.id = instance_count++;
```
nemanja-boric-sociomantic: `this.id = typeof(this).instance_count++` or similar? Do we require `this.` on the static members?

gavin-norman-sociomantic: Updated.

nemanja-boric-sociomantic (approved):

PR #152, opened 2017-10-16
Fix resuming terminated fiber in all-nodes request

david-eckardt-sociomantic:
In the client `Request.resumeSuspendedHandlers` can attempt to resume a request handler fiber of an all-nodes request that has terminated already, causing an assertion to fail. Fix this by resuming only fibers that are in `HOLD` state rather than those that are not in `EXEC` state.


gavin-norman-sociomantic (approved):
Ooh interesting! Good catch.

PR #161, opened 2017-11-01
Add missing import

mathias-lang-sociomantic:


gavin-norman-sociomantic:
Could you make a note in the commit message explaining why this change is required? (Presumably it's because of changes in dmd version `<something>`.)

mathias-lang-sociomantic:
Updated

PR #162, opened 2017-11-01
Fix usage of formatter in ProtocolError

mathias-lang-sociomantic:
The code assumed that casting a delegate was safe and that the compiler would forgive you for returning a value from a void-returning delegate.
But it didn't.

### DIFF
--- a/src/swarm/neo/protocol/ProtocolError.d
+++ b/src/swarm/neo/protocol/ProtocolError.d
@@ -29,9 +29,6 @@ class ProtocolError: Exception
     import ocean.transition;
     import ocean.text.convert.Formatter;
 
-    static if (!is(FormatterSink))
-        alias Sink FormatterSink;
-
     /***************************************************************************
 
         Sets exception information for this instance.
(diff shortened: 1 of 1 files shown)


src/swarm/neo/protocol/ProtocolError.d
```diff
@@ -96,17 +96,20 @@ class ProtocolError: Exception
...
         this.reused_msg.reset();
-        sformat(
-            cast(FormatterSink) (cstring chunk) {
-                this.reused_msg ~= chunk;
-                return chunk.length;
-            },
-            fmt,
-            args
-        );
+        static if (is(FormatterSink))
```
gavin-norman-sociomantic: Does this depend on a certain ocean version?

gavin-norman-sociomantic:
Maybe note which compiler version this is talking about. It clearly compiles without the current "mainstream" compiler complaining.

src/swarm/neo/protocol/ProtocolError.d
mathias-lang-sociomantic: Yes, `FormatterSink` was introduced with the deprecation of the `size_t` overload.

gavin-norman-sociomantic: I see. I was wondering if it might be clearer to statically check for the library version.

mathias-lang-sociomantic: I can't do that, because `LibFeature` wasn't properly updated at the time (that was found when I tried to do a similar change in another library).

gavin-norman-sociomantic: Ah ok, fair enough.

That makes me wonder: do we have a policy for removing lib-version based `static if`s at some point? We probably should. It'd be nice to make sure (via a comment?) that this conditional gets removed during such a purge.

mathias-lang-sociomantic:
Updated.

src/swarm/neo/protocol/ProtocolError.d
gavin-norman-sociomantic: What do you think about this?

mathias-lang-sociomantic: > That makes me wonder: do we have a policy for removing lib-version based static ifs at some point? We probably should. It'd be nice to make sure (via a comment?) that this conditional gets removed during such a purge.

We don't at the moment, because no one have been using them before, so the issue hasn't come up. I added a comment (and make sure to include the word "deprecated" so grep will show it).

gavin-norman-sociomantic: :+1: I missed the extra comment before.

mathias-lang-sociomantic: No you didn't, I missed your comment :P

PR #169, opened 2017-11-08
Store name of connected client and make accessible at the level of request handlers

gavin-norman-sociomantic:


src/swarm/neo/protocol/connect/NodeConnect.d
```diff
@@ -162,21 +162,29 @@ class NodeConnect
         {
             try
             {
-                enforce(this.e_auth_rejected.set("Empty client name"),          // Validate the client name.
+                // Validate the client name.
+                enforce(this.e_auth_rejected.set("Empty client name"),
```
nemanja-boric-sociomantic, line 170: Not related to this change, but there's something really weird in the arguments order of these `enforce` calls.

nemanja-boric-sociomantic (approved):

gavin-norman-sociomantic:
I might be able to add a test to check that the correct client name is stored. One sec.

gavin-norman-sociomantic:
Actually there's not a convenient way of doing it. (I could add a new request -- `CheckClientName` -- to the test node/client, but that's too much hassle for such a minor point.)

src/swarm/neo/protocol/connect/NodeConnect.d
```diff
@@ -185,6 +189,7 @@ class NodeConnect
 
                     // Authentication successful.
                     success = true;
+                    this.client_name.copy(client_name);
```
nemanja-boric-sociomantic: Should you clean `this.client_name` somewhere? Not sure how feasible this scenario is:

1. Old client connects and authenticates
2. The client disconnects
3. The `this.connected_client` is called, it returns the old client, but this client is no longer connected.

gavin-norman-sociomantic: Good point. It's obscure, but probably worth shoring up. I'll try to find a way.

gavin-norman-sociomantic: Updated.

relnotes/client-name.feature.md
```diff
@@ -0,0 +1,8 @@
+## Name of connected client is now accessible via `Connection`
+
+`swarm.neo.node.Connection`
+
+The node `Connection` class has a new method -- `connected_client` -- that
+returns the name of the connected client received during authentication. This
+is useful for logging.
+
```
nemanja-boric-sociomantic: I assume the commit order on the github is wrong (because this file in the last gh commit is changed in the first gh commit). Useful trick: `git rebase -i -x 'git commit --amend --date=$(date +%s) --no-edit' upstream/v4.x.x` will fix it. Not important for this PR, as it's trivial.

Another small point: we don't use indentation in the new release notes format anymore (IIRC we used to have an asterisk in front of the module names and the description would be indented by two spaces)?

gavin-norman-sociomantic: Yeah the commit order is displayed wrong.

>we don't use indentation in the new release notes format anymore?

Correct. We're using headings instead of a bullet list.

nemanja-boric-sociomantic: So, this is wrong: https://github.com/sociomantic-tsunami/ocean/blob/186faa8572df58714fa54b0856a8ffcaf21b41d2/relnotes/baguette.migration.md ?

gavin-norman-sociomantic: According to https://github.com/sociomantic-tsunami/neptune/blob/master/doc/library-contributor.rst, yes.

gavin-norman-sociomantic: (Though it's only _slightly_ wrong.)

nemanja-boric-sociomantic: Thanks!

nemanja-boric-sociomantic (approved):

PR #178, opened 2017-11-15
Clean up request context serialization

gavin-norman-sociomantic:
Work in progress.

### DIFF
--- a/src/swarm/neo/client/mixins/RequestCore.d
+++ b/src/swarm/neo/client/mixins/RequestCore.d
@@ -259,11 +259,42 @@ public template RequestCore ( RequestType request_type_, ubyte request_code,
 
         /***********************************************************************
 
-            Object for acquiring resources required by the request.
+            Object for acquiring shared resources required by the request.
 
         ***********************************************************************/
 
-        public SerializableReferenceType!(Object) request_resources;
+        public Object shared_resources;
+
+        /***********************************************************************
+
+            Note: this struct is essentially deprecated. It is only needed in
+            order to maintain the public API of Context. In the next major
+            release, this wrapper struct can be removed.
+
+        ***********************************************************************/
+
+        private struct RequestResources
+        {
+            private Object* request_resources;
+
+            public Object get ( )
+            {
+                return *this.request_resources;
+            }
+        }
+
+        /***********************************************************************
+
+            Returns:
+                instance to shared resources wrapper struct
+
+        ***********************************************************************/
+
+        deprecated("Access `shared_resources` object directly instead.")
+        public RequestResources request_resources ( )
+        {
+            return RequestResources(&this.shared_resources);
+        }
 
         /***********************************************************************
 
--- a/src/swarm/neo/client/mixins/RequestCore.d
+++ b/src/swarm/neo/client/mixins/RequestCore.d
@@ -251,14 +251,6 @@ public template RequestCore ( RequestType request_type_, ubyte request_code,
 
         /***********************************************************************
 
-            Global working data required by the request.
-
-        ***********************************************************************/
-
-        public SharedWorking shared_working;
-
-        /***********************************************************************
-
             Object for acquiring shared resources required by the request.
 
         ***********************************************************************/
@@ -298,36 +290,20 @@ public template RequestCore ( RequestType request_type_, ubyte request_code,
 
         /***********************************************************************
 
+            Global working data required by the request.
+
+        ***********************************************************************/
+
+        public SharedWorking shared_working;
+
+        /***********************************************************************
+
             The ID of the request in the request set (set by RequestSet, passed
             to the notifier).
 
         ***********************************************************************/
 
         public RequestId request_id;
-
-        /***********************************************************************
-
-            Helper function to properly serialize user-specified params for the
-            request into the `user_params` field. An intermediary serialzation
-            buffer is used to safely convert from the const params provided by
-            the user to the mutable params required internally (in order to be
-            able to deserialize into a Context instance).
-
-            Params:
-                params = user-specified request parameters to store
-                serialize_buffer = intermediary serialization buffer
-
-        ***********************************************************************/
-
-        public void setUserSpecifiedParams ( Const!(UserSpecifiedParams) params,
-            ref ubyte[] serialize_buffer )
-        {
-            auto serialized = Serializer.serialize(params, serialize_buffer);
-            enableStomping(serialized);
-            auto deserialized = Deserializer.deserialize!(UserSpecifiedParams)(
-                serialized);
-            this.user_params = *deserialized.ptr;
-        }
     }
 
     /***************************************************************************
--- a/src/swarm/neo/client/mixins/RequestCore.d
+++ b/src/swarm/neo/client/mixins/RequestCore.d
@@ -121,6 +121,7 @@ public template RequestCore ( RequestType request_type_, ubyte request_code,
     import swarm.neo.request.Command;
     import swarm.neo.AddrPort;
     import swarm.neo.client.NotifierTypes;
+    import swarm.neo.util.StructPacker;
 
     /***************************************************************************
 
@@ -369,7 +370,7 @@ public template RequestCore ( RequestType request_type_, ubyte request_code,
 
     /***************************************************************************
 
-        Private helper function to convert the raw, serialized context into a
+        Private helper function to convert the raw, packed context into a
         Context instance.
 
         Params:
@@ -382,7 +383,7 @@ public template RequestCore ( RequestType request_type_, ubyte request_code,
 
     private static Context* getContext ( void[] context_blob )
     {
-        return Contiguous!(Context)(context_blob).ptr;
+        return unpack!(Context)(context_blob);
     }
 
     /***************************************************************************
--- a/src/swarm/neo/util/StructPacker.d
+++ b/src/swarm/neo/util/StructPacker.d
@@ -1,0 +1,380 @@
+/*******************************************************************************
+
+    Functions to pack a struct instance and all contained array buffers into a
+    contiguous chunk of memory.
+
+    The packer is similar to ocean's contiguous serializer, but:
+        * Can handle non-array reference types (e.g. delegates, pointers).
+        * Can handle unions.
+        * Automatically sets the pointers of packed arrays to point to the
+          array data contained in the packed buffer.
+
+    Copyright:
+        Copyright (c) 2017 sociomantic labs GmbH. All rights reserved
+
+    License:
+        Boost Software License Version 1.0. See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module swarm.neo.util.StructPacker;
+
+import ocean.transition;
+import ocean.core.Traits;
+
+/*******************************************************************************
+
+    Packs `s` into the provided buffer, ensuring that the content of all
+    dynamic array fields are contained fully within the buffer.
+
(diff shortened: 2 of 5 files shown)


src/swarm/neo/util/StructSerializer.d
```diff
@@ -0,0 +1,335 @@
...
+    enableStomping(buf);
+    buf[] = (cast(void*)&s)[0..buf.length];
+
+    // Handle any array fields.
+    auto serialized_struct = deserialize!(S)(buf);
+    foreach ( i, ref f; (*serialized_struct).tupleof )
+    {
+        static if ( isDynamicArrayType!(typeof(f)) )
+            serializeDynamicArray(f, buf);
+    }
```
david-eckardt-sociomantic: Alternative, slightly shorter:
```d
// Do a simple, binary copy of the struct.
buf.length = S.sizeof;
enableStomping(buf);
auto serialized_struct = cast(S*)buf.ptr;
*serialized_struct = s;
// Handle any array fields.
foreach ( i, ref f; (*serialized_struct).tupleof )
// ...
```

gavin-norman-sociomantic: Ah yeah, good idea.

gavin-norman-sociomantic:
Gah damn, I just noticed that this restriction: https://github.com/sociomantic-tsunami/swarm/pull/178/files#diff-4195e2bb4e0d50489a9738a2c43711a1R199 makes this serializer unsuitable for the task it's intended for. I was thinking we're serializing the args struct containing the arguments provided by the user. That struct is actually nested inside two others, though: https://github.com/sociomantic-tsunami/swarm/blob/v4.x.x/src/swarm/neo/client/mixins/RequestCore.d#L179 and https://github.com/sociomantic-tsunami/swarm/blob/v4.x.x/src/swarm/neo/client/mixins/RequestCore.d#L249.

Might need to rethink this...

gavin-norman-sociomantic:
Updated with support for arrays inside nested structs.

Next step: adapt internal request context serialization to use this.

gavin-norman-sociomantic:
Updated: fixed D2 errors.

src/swarm/neo/util/StructSerializer.d
```diff
@@ -0,0 +1,347 @@
...
+        A = type of array to serialize
+        array = slice to serialize and update to point to the serialized content
+        buf = buffer to serialize into
+
+*******************************************************************************/
+
+private void serializeDynamicArray ( A ) ( ref A array, ref void[] buf )
+{
+    static assert(isDynamicArrayType!(A));
+    static assert(isPrimitiveType!(BaseTypeOfArrays!(A)));
```
david-eckardt-sociomantic: Suggest
- using `A` as the array *element* type (unless is causes some type qualifier related complication) and
- disallowing indirections other than arrays,  i.e. also allowing struct/union/static array elements as long as they are plain values, instead of allowing elements of primitive type only.
```d
private void serializeDynamicArray ( A ) ( ref A[] array, ref void[] buf )
{
    // ...

    // Recursively serialize arrays of arrays.
    static if (isDynamicArrayType!(A))
    {
        foreach ( ref e; array )
            serializeDynamicArray(e, buf);
    }
    else
        static assert(!hasIndirections!(A));
}
```

gavin-norman-sociomantic:
Updated: adapt swarm client's internals to use the new serializer.

This adaptation contains at least one breaking change, so I'll have to consider whether to try to do as much as I can without the breaking change or whether to just move this PR to v5.0.0.

src/swarm/neo/util/StructSerializer.d
```diff
@@ -0,0 +1,349 @@
...
+}
+
+version ( UnitTest )
+{
+    import ocean.core.Test;
+    import ocean.core.DeepCompare : deepEquals;
+
+    // Test helper function to serialize, deserialize, and compare against the
+    // original.
+    bool serialzeDeserialize ( S ) ( S s )
```
mihails-strasuns-sociomantic: typoe `serialze`

src/swarm/neo/util/StructSerializer.d
```diff
@@ -0,0 +1,349 @@
...
+        S = type to deserialize
+        serialized = buffer containing the serialized thing
+
+    Returns:
+        pointer to the buffer interpreted as a pointer to an instance of the
+        serialized type
+
+*******************************************************************************/
+
+public S* deserialize ( S ) ( Const!(void)[] serialized )
```
david-eckardt-sociomantic: This casts the `const` away from `serialized`. It looks like a case for `inout`,
```d
Inout!(S)* deserialize ( S ) ( Inout!(void)[] serialized )
```
so that the qualifier of the return type follows that of the argument.

src/swarm/neo/util/StructSerializer.d
```diff
@@ -0,0 +1,349 @@
...
+    content just added to the serialization buffer.
+
+    Params:
+        S = type of struct to serialize dynamic arrays from
+        s = struct instance to serialize dynamic arrays from
+        buf = buffer to serialize into
+
+*******************************************************************************/
+
+private void seriaizeStructDynamicArrays ( S ) ( ref S s, ref void[] buf )
```
david-eckardt-sociomantic: Typo `seriaize`

src/swarm/neo/util/StructSerializer.d
gavin-norman-sociomantic: Mm I don't see why that would be useful...?

1. `Const!(void)[] serialized` states that we never modify the input buffer.
2. `S*` states that we return a pointer to some type. (Note that the user can do `auto s = deserialize!(Const!(Whatever))(buffer);` if they want to.)

gavin-norman-sociomantic:
Updated:
* `deserialize` function now accepts a mutable buffer. This is perfectly fine for our usage (we always deserialize from a mutable buffer) and prevents the compiler from making any assumptions about the buffer potentially being immutable.

gavin-norman-sociomantic:
Updated:
* Renamed to un/packer, as suggested by Mihails.
* Simplified `serializeDynamicArray`, as suggested by David.

src/swarm/neo/util/StructPacker.d
```diff
@@ -0,0 +1,349 @@
+/*******************************************************************************
+
+    Functions to pack a struct instance and all contained array buffers into a
+    contiguous chunk of memory.
```
mihails-strasuns-sociomantic, line 4: I would explicitly stress that packed data can't be copied/moved because it contains interior pointers.

gavin-norman-sociomantic:
Updated: replaced breaking change with deprecation path.

gavin-norman-sociomantic:
Updated:
* Tested with dhtproto.
* Added comment as suggested by Mihails.
* Unblocked.

gavin-norman-sociomantic:
Ah... release notes coming up.

gavin-norman-sociomantic:
Updated with release notes for the deprecation. I decided not to add release notes for the new feature (the un/packer functions), as they're purely internal to the client and not intended to be used externally.

src/swarm/neo/util/StructPacker.d
```diff
@@ -0,0 +1,354 @@
...
+
+    Copyright:
+        Copyright (c) 2016-2017 sociomantic labs GmbH. All rights reserved
```
nemanja-boric-sociomantic: Check copyright (only 2017 needed).

gavin-norman-sociomantic: Updated.

src/swarm/neo/util/StructPacker.d
gavin-norman-sociomantic: Mentioned in method doc.

mihails-strasuns-sociomantic: Better to also put it into the module doc as it is the main design point of the whole concept, immediately answering "how is it different from serializer" question.

gavin-norman-sociomantic: :+1: That's a minor difference, I'd say. I'll add something, anyway.

gavin-norman-sociomantic: Updated.

gavin-norman-sociomantic:
Any further comments on this?

src/swarm/neo/client/mixins/RequestCore.d
```diff
@@ -259,11 +259,42 @@ public template RequestCore ( RequestType request_type_, ubyte request_code,
...
+
+            Note: this struct is essentially deprecated. It is only needed in
+            order to maintain the public API of Context. In the next major
+            release, this wrapper struct can be removed.
+
+        ***********************************************************************/
+
+        private struct RequestResources
+        {
+            private Object* request_resources;
```
david-eckardt-sociomantic: Why is this a pointer?

gavin-norman-sociomantic: So that the return value of `get` always reflects the value of `shared_resources` in the outer context.

gavin-norman-sociomantic: It's a bit theoretical, but you could do:
```D
auto rr = context.request_resources();
do_something(rr.get());
context.shared_resources = something_else;
do_something(rr.get()); // should reflect the changed value
```

david-eckardt-sociomantic: I see. Fair enough.

src/swarm/neo/util/StructPacker.d
```diff
@@ -0,0 +1,360 @@
...
+    bool error;
+
+    foreach ( i, F; typeof(S.tupleof) )
+    {
+        // Arrays are allowed, if...
+        static if ( isDynamicArrayType!(F) )
+        {
+            // ...they only contain primitive types.
+            static if ( !isPrimitiveType!(BaseTypeOfArrays!(F)) )
+                error = true;
```
david-eckardt-sociomantic: - For the sake of an error message that is more comprehensive than "something is wrong" I suggest using `static assert`s on the error conditions with the unsupported type and maybe even the field name in the message, and naming the function `void checkPackable` or something similar.
- Why are dynamic arrays of pointers/delegates/objects etc. disallowed, again, while static arrays of such types are allowed?

src/swarm/neo/util/StructPacker.d
```diff
@@ -0,0 +1,360 @@
...
+        }
+
+        // Structs are allowed, if...
+        static if ( is (F == struct) )
+        {
+            // ...they could be packed on their own.
+            static if ( !canPack!(F) )
+                error = true;
+        }
```
david-eckardt-sociomantic: Not sure if it is that important, but you may want to check unions and allow only those that don't contain dynamic arrays.

src/swarm/neo/util/StructPacker.d
gavin-norman-sociomantic: >Why are dynamic arrays of pointers/delegates/objects etc. disallowed, again, while static arrays of such types are allowed?

Because dynamic array contents are copied to the end of the packed buffer and may thus become misaligned.

david-eckardt-sociomantic: Ah, right. Now I remember.

gavin-norman-sociomantic: I'll add a comment explaining this.

gavin-norman-sociomantic:
Updated.

david-eckardt-sociomantic (approved):

PR #180, opened 2017-11-20
Switch from Volkswagen testing to BMW testing

mathias-lang-sociomantic:
@mihails-strasuns-sociomantic : Whie compiling, I get the following warning:
`./src/swarm/Const.d(112): Warning: toHash() must be declared as extern (D) size_t toHash() const nothrow @safe, not const ulong()`

I see two problems here:
- `d1to2fix` will forbid those tokens.
- Compilation passed even with `-w` being passed on the command line

### DIFF
--- a/src/swarm/Const.d
+++ b/src/swarm/Const.d
@@ -109,7 +109,7 @@ public struct NodeItem
 
     ***************************************************************************/
 
-    public hash_t toHash ( )
+    public hash_t toHash ( ) /* d1to2fix_inject: const */
     {
         return Fnv1a(this.Address) ^ this.Port;
     }
--- a/integrationtest/client_auth/main.d
+++ b/integrationtest/client_auth/main.d
@@ -10,9 +10,9 @@
 
 *******************************************************************************/
 
-module test.client_auth.main;
+module integrationtest.client_auth.main;
 
-import test.neo.client.Client;
+import integrationtest.neo.client.Client;
 
 import swarm.neo.authentication.CredentialsFile;
 
@@ -24,6 +24,8 @@ import ocean.io.select.EpollSelectDispatcher;
 import Config = ocean.util.config.ConfigFiller;
 import ocean.util.config.ConfigParser;
 
+version (UnitTest) {}
+else
 void main ( )
 {
     // Create temporary sandbox directory to write files to.
--- a/integrationtest/neo/client/Client.d
+++ b/integrationtest/neo/client/Client.d
@@ -15,7 +15,7 @@
 
 *******************************************************************************/
 
-module test.neo.client.Client;
+module integrationtest.neo.client.Client;
 
 import ocean.transition;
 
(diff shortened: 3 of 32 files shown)


src/swarm/Const.d
```diff
@@ -109,7 +109,7 @@ public struct NodeItem
```
nemanja-boric-sociomantic, line 109: Is the commit message right (`Since HEAD is currently using v5.x.x`)?

mathias-lang-sociomantic: No you're right. Fixed.

mathias-lang-sociomantic:
Added 2 bonus commits since I'm at it

mihails-strasuns-sociomantic:
> Warning: toHash() must be declared as extern (D) size_t toHash() const nothrow @safe, not const ulong()

This is going to be a problem - we have enough trouble with `const`/`immutable` forced on us by runtime, caring about `@safe` is definitely not desired. I would even consider.

Does warning appear on 2.071 or later version?

gavin-norman-sociomantic:
>I would even consider.

What would you consider?

mihails-strasuns-sociomantic:
> What would you consider?

Contacting upstream to find a way to address this situation for code bases that don't want `@safe` :)

mathias-lang-sociomantic:
> Does warning appear on 2.071 or later version?

Appears with `dmd-transitional` :(

mathias-lang-sociomantic:
Although [according to the code](https://github.com/dlang/dmd/blame/v2.071.0/src/todt.c#L1283-L1295) we can "safely" ignore this warning.

mihails-strasuns-sociomantic:
Can you turn submodule upgrade commit into fixup for sociomantic-tsunami/swarm@60156cc ?

I would ignore warning for now but keep problem in mind :(

mihails-strasuns-sociomantic:
And source fixup should probably go as fixup for https://github.com/sociomantic-tsunami/swarm/commit/3c933052abec1a9aec87a0f21c7d3fab7d2e913a

mathias-lang-sociomantic:
So basically all fixup ? :P

mihails-strasuns-sociomantic:
Well, it is one big fixup PR, isn't it? :)

mathias-lang-sociomantic:
Like this ?

mihails-strasuns-sociomantic:
Hm, you seem to have put it all into single commit now - is that intentional?

mathias-lang-sociomantic:
Only the first 2, which fixes the submodule update commit. They don't make sense separated if you use them to fixup.

mihails-strasuns-sociomantic:
Some changes look related to compiler upgrade to 2.071, which was a separate commit. I notice one import  adjustment for example. False match?

mathias-lang-sociomantic:
Hum I forgot I am using v2.071 by default now. I moved the protection change to its own commit. For the `const` change I *think* they belong here anyway.

mihails-strasuns-sociomantic (approved):
Thanks!

mihails-strasuns-sociomantic:
@gavin-norman-sociomantic FYI I will make another swarm v5.x.x rebase myself after tagging v4.0.0 alpha in ocean this week, will double-check all tests at that point.

mihails-strasuns-sociomantic:
Updated swarm v5.x.x to use ocean `v4.0.0-alpha.0`, squashed fixup commits and force pushed again.

PR #184, opened 2017-11-22
Optimise array handling of RequestEventDispatcher

gavin-norman-sociomantic:


src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -701,6 +681,27 @@ public struct RequestEventDispatcher
...
+        Unregisters all events for `fiber`.
+
+        Params:
+            fiber = fiber to unregister events for
+
+    ***************************************************************************/
+
+    private void unregisterFiber ( MessageFiber fiber )
+    {
+        this.waiting_fibers.length = Array.moveToEnd(this.waiting_fibers,
```
nemanja-boric-sociomantic: `removeShift` is not suitable here?

gavin-norman-sociomantic: `removeShift` removes a single element, whereas here we want to remove several matching elements in one pass.

gavin-norman-sociomantic: (The 2nd commit refactors this method again, anyway.)

nemanja-boric-sociomantic: > removeShift removes a single element, whereas here we want to remove several matching elements in one pass.

Ah :+1:

gavin-norman-sociomantic:
>Refactor RequestEventDispatcher to never modify length of internal array

I suppose this should say "never _reduce_ length", actually.

gavin-norman-sociomantic:
Updated: added protection attribute; fixed commit message.

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -629,39 +634,56 @@ public struct RequestEventDispatcher
...
+            if ( waiting_fiber.fiber == fiber && waiting_fiber.event == event )
+            {
+                assert(in_list is null);
+                assert(!waiting_fiber.enabled);
+                in_list = &waiting_fiber;
+                break;
+            }
+
+            // Ensure the user isn't trying to do anythign crazy.
+            if ( waiting_fiber.enabled &&
```
gavin-norman-sociomantic: Hmm I notice that the main loop to find the matching element can `break`, which means that the sanity check loop doesn't complete. These should probably be two separate loops now.

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -251,9 +251,19 @@ public struct RequestEventDispatcher
 
         /// Event being awaited.
         EventRegistration event;
+
+        /// Is this instance currently enabled? If false, it will be ignored
+        /// when dispatching events.
+        bool enabled;
```
david-eckardt-sociomantic, line 257: Could you use `event.active.none` (i.e. `!event.active`) to indicate that this instance is disabled?

src/swarm/neo/request/RequestEventDispatcher.d
david-eckardt-sociomantic: Or don't `break` and set `in_list` only if it is not `null`.

gavin-norman-sociomantic: Updated.

src/swarm/neo/request/RequestEventDispatcher.d
gavin-norman-sociomantic: Unfortunately not, as the loops in `register` and `unregister` match the fiber/event pair.

src/swarm/neo/request/RequestEventDispatcher.d
gavin-norman-sociomantic: Updated with David's suggestion.

david-eckardt-sociomantic: Is `assert(in_list is null)` still correct now that this loop doesn't `break`?

gavin-norman-sociomantic: Yes, because it should be impossible to register the same fiber/event pair twice.

gavin-norman-sociomantic: (That assertion actually came from an earlier version of the code which didn't break the loop.)

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -629,39 +634,56 @@ public struct RequestEventDispatcher
...
+            }
+
+            // Ensure the user isn't trying to do anything crazy.
+            if ( waiting_fiber.enabled &&
+                waiting_fiber.event.active == event.active )
+            {
+                with ( event.Active ) switch ( event.active )
+                {
+                    case message:
                         enforce(waiting_fiber.event.message.type
```
nemanja-boric-sociomantic: This is something else, not related to this commit, but shouldn't this be `verify` ?

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -756,11 +783,12 @@ public struct RequestEventDispatcher
     }
     body
     {
-        foreach ( i, waiting_fiber; this.waiting_fibers )
-            if ( waiting_fiber.event.active == waiting_fiber.event.active.send )
+        foreach ( ref waiting_fiber; this.waiting_fibers )
+            if ( waiting_fiber.enabled &&
+                waiting_fiber.event.active == waiting_fiber.event.active.send )
             {
                 WaitingFiber ret = waiting_fiber;
```
nemanja-boric-sociomantic:
```
this.disable(waiting_fiber);
return waiting_fiber;
```

or am I missing something?

src/swarm/neo/request/RequestEventDispatcher.d
gavin-norman-sociomantic: I suppose you're right. It would be a programmer error, not an error based on processing run-time data. `verify` doesn't exist in the ocean version that swarm uses, but these could be turned into `assert`s.

nemanja-boric-sociomantic: Maybe just add a comment here, so we know when we `verify` everything?

src/swarm/neo/request/RequestEventDispatcher.d
gavin-norman-sociomantic: I think the idea is to make a copy of it before it's disabled, then return the copy.

gavin-norman-sociomantic: (I don't remember what difference this could make, though.)

nemanja-boric-sociomantic: Ah, could it be due the fact that the old code was moving this instance around the array, invalidating this reference to it? In that case, it's not needed any more.

gavin-norman-sociomantic: Maybe. I'll have a look at the usage.

gavin-norman-sociomantic: Yep, you're right. Looks like a refactoring omission. Updated.

src/swarm/neo/request/RequestEventDispatcher.d
nemanja-boric-sociomantic: Ping here, and LGTM.

nemanja-boric-sociomantic: Two things: 

- `enforce`` <- two ticks at the end
- maybe add `should be asserts (and ideally verify)` so we can `grep` for `assert` when do the transition.

gavin-norman-sociomantic: Ah yes. Updated.

nemanja-boric-sociomantic (approved):

src/swarm/neo/request/RequestEventDispatcher.d
gavin-norman-sociomantic: Oh yeah. Updated. (The comments already contain the string "assert".)

nemanja-boric-sociomantic: Yes, I just saw it, sorry!

PR #188, opened 2017-12-05
Add new record batcher

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Preview of rough implementation, for early feedback on API etc.

gavin-norman-sociomantic:
Ah, build failure due to dmd1 not understanding `final switch`. But anyway, the PR is just for early feedback right now.

src/swarm/neo/util/Batch.d
```diff
@@ -0,0 +1,236 @@
...
+    public bool add ( T ... ) ( T items, void delegate ( ) send_batch )
+    {
+        if ( !this.fits(items) )
+            return false;
+
+        foreach ( item; items )
+            this.addItem(item);
+
+        if ( this.buffer.length >= this.max_size )
+            send_batch();
```
david-eckardt-sociomantic: This looks wrong: `send_batch` is called only if `items` precisely fit in the buffer, otherwise one of the `if` conditions yields `false`.

gavin-norman-sociomantic: Mm, well it's intended to only call `send_batch` if the buffer is full. Seems correct to me...?

gavin-norman-sociomantic:
Added an extra commit (to be squashed) to clear the batch after sending.

src/swarm/neo/util/Batch.d
david-eckardt-sociomantic: Suppose `this.max_size == 60`, and you call `add` repeatedly, each time with one item of size 40. On the first call `this.buffer.length >= this.max_size` yields `false` so `send_batch` is not called. On subsequent calls `this.fits(items)` returns `false` so `send_batch` is not called either. So `send_batch` is never called.

gavin-norman-sociomantic: You're misreading `this.fits`.

david-eckardt-sociomantic: Oh, I see. Sorry.

david-eckardt-sociomantic: LGTM.

david-eckardt-sociomantic: Suggest raising an error (`assert/verify`) if `this.buffer.length >= this.max_size` and this method is called, or calling `clear` after `send_batch`.

gavin-norman-sociomantic: Yep, I added an extra commit that clears after sending :+1:

src/swarm/neo/util/Batch.d
```diff
@@ -0,0 +1,236 @@
...
+        static if ( isDynamicArrayType!(T) )
+        {
+            static assert(rankOfArray!(T) == 1,
+                "Only 1d arrays and primitive types are supported, not "
+                ~ T.stringof);
+            return size_t.sizeof + (item.length * ElementTypeOfArray!(T).sizeof);
+        }
+        else
+        {
+            static assert(isPrimitiveType!(T),
```
david-eckardt-sociomantic: `static assert(!hasIndirections!(T))` allows all value types.

src/swarm/neo/util/Batch.d
```diff
@@ -0,0 +1,236 @@
...
+    {
+        return *this.buffer;
+    }
+
+    public Const!(void)[] getCompressed ( Lzo lzo, ref void[] compress_buf )
+    {
+        // Set destination to max possible length.
+        compress_buf.length =
+            lzo.maxCompressedLength(this.buffer.length) + size_t.sizeof;
+        enableStomping(compress_buf);
```
david-eckardt-sociomantic, line 178: I was told `enableStomping` is expensive. Since it needs to be called only before increasing the buffer length, you could call it only in that case. Since LZO normally compresses the data one `enableStomping` call can usually be avoided.

src/swarm/neo/util/Batch.d
```diff
@@ -0,0 +1,239 @@
...
+                t.test(!sent);
+                break;
+
+            case Sent:
+                t.test(added);
+                t.test!("==")(batch.get.length, expected_length);
+                t.test(sent);
+                break;
+
+            default: assert(false);
```
nemanja-boric-sociomantic, line 485: Must be guarded by `version(D_Version2){} else { ... }`

gavin-norman-sociomantic: Yeah, actually as this is just for a unittest, I'll remove the `final`.

src/swarm/neo/util/Batch.d
```diff
@@ -0,0 +1,239 @@
...
+    {
+        if ( !this.fits(items) )
+            return false;
+
+        foreach ( item; items )
+            this.addItem(item);
+
+        if ( this.buffer.length >= this.max_size )
+        {
+            send_batch();
```
nemanja-boric-sociomantic: Perhaps `send_batch` could accept the `Const!(void)[]` of the compressed data, as we're now clearing the batch after sending.

src/swarm/neo/util/Batch.d
```diff
@@ -0,0 +1,239 @@
...
+
+        return true;
+    }
+
+    public Const!(void)[] get ( )
+    {
+        return *this.buffer;
+    }
+
+    public Const!(void)[] getCompressed ( Lzo lzo, ref void[] compress_buf )
```
nemanja-boric-sociomantic: There's a bit of asymmetry here as the RecordBatcher encapsulates the way how the compressed batch is packed (size + data). Perhaps, having the (static?) method to uncompress it into a buffer will come in handy?

src/swarm/neo/util/Batch.d
gavin-norman-sociomantic: The reason for not passing it is so the user can choose to call `get()` or `getCompressed()`, as they require.

nemanja-boric-sociomantic: Fair enough :+1:

src/swarm/neo/util/Batch.d
gavin-norman-sociomantic: Asymmetry with what?

nemanja-boric-sociomantic: With the encapsulation. Knowing how to unpack the compressed batch, user of the record batcher needs to know the internals of the RecordBatcher. Maybe it's easier to just send away ` struct CompressedBatch { ssize_t uncompressed_length; Const!(void)[] data; }`.

gavin-norman-sociomantic:
Added another commit with the record batch reader.

src/swarm/neo/util/Batch.d
gavin-norman-sociomantic: Ah no, the record batch reader knows how to do this. See new commit.

src/swarm/neo/util/Batch.d
```diff
@@ -148,9 +148,78 @@ struct BatchWriter
...
+                this.extractItem(item);
+
+            if ( auto r = dg(items) )
+                return r;
+        }
+
+        return 0;
+    }
+
+    private void extractItem ( T ) ( ref T item )
```
nemanja-boric-sociomantic: Hmm, I would argue this should be `public` and probably `opApply` removed. The reason is that you can pack key and value of the distinct types, but you can't extract them with the `opApply`.

src/swarm/neo/util/Batch.d
nemanja-boric-sociomantic: :+1: I wasn't aware of this.

src/swarm/neo/util/Batch.d
gavin-norman-sociomantic: Yes you can. See the unittest.

nemanja-boric-sociomantic: If you're referring to the last unittest, that's just packing a struct of the `struct S { key; value; }` I was thinking about this:

Node:
```
time_t key; void[] value;
batch.add(key, value, &sendBatch);
```

How you extract this? Because, you can pack it.

gavin-norman-sociomantic: It looks a bit deceptive. Look at the lines where it [adds to the batch](https://github.com/sociomantic-tsunami/swarm/pull/188/files#diff-f9e677247e1ecb227ec79f3aebc4a829R379) and [extracts from the batch](https://github.com/sociomantic-tsunami/swarm/pull/188/files#diff-f9e677247e1ecb227ec79f3aebc4a829R384).

gavin-norman-sociomantic: (The struct is only used as a convenient way of making a list of "records" to be put in the batch.)

nemanja-boric-sociomantic: Hmm, all right. Though I think it looks weird, but I'll not complain. Would you be against adding `public bool extractItems(T...)(ref T items)`?

gavin-norman-sociomantic: >Would you be against adding public bool extractItems(T...)(ref T items)?

To extract just a single tuple / record? I wondered if there's a use case for that.

>Though I think it looks weird, but I'll not complain.

If you think the API looks weird, please complain. If you think the unittest looks weird, I'm not so bothered.

nemanja-boric-sociomantic: I think I'm missing something here. Can you use `opApply` to extract the tuple or arbitrary degree?

> To extract just a single tuple / record?

To extract `(a1, a2, a3, ..., an)` (same as you can add them using `Batch.add` method)

gavin-norman-sociomantic: Yes, I guess you're missing something ;) This is what the usage looks like:

Node-side:
```D
hash_t key;
mstring value;
batch.add(key, value);
```
Client-side:
```D
foreach ( key, value; new BatchReader!(hash_t, cstring)(batch) )
```

nemanja-boric-sociomantic: For the record, I was missing the template `BatchReader` argument.

gavin-norman-sociomantic:
Changed `BatchWriter` to a template and squashed.

src/swarm/neo/util/Batch.d
```diff
@@ -0,0 +1,399 @@
...
+    {
+        hash_t key;
+        mstring value;
+    }
+    KeyValue[] values = [KeyValue(12, "hi"), KeyValue(23, "bye"),
+        KeyValue(34, "whatever")];
+
+    {
+        const max_len = 256;
+        BatchWriter!(hash_t, cstring) batch;
```
nemanja-boric-sociomantic: Suggestion `BatchWritter!(typeof(KeyValue.tupleof))`

src/swarm/neo/util/Batch.d
david-eckardt-sociomantic: Suggest changing the name of the method template parameter to `Item` or something to avoid confusion with the class template parameter.

src/swarm/neo/util/Batch.d
```diff
@@ -148,9 +148,78 @@ struct BatchWriter
...
+                ~ T.stringof);
+            auto slice = (cast(void*)&item)[0..T.sizeof];
+            slice[] = this.extractBytes(T.sizeof)[];
+        }
+    }
+
+    private Const!(void)[] extractBytes ( size_t bytes )
+    in
+    {
+        assert(this.remaining.length >= bytes);
```
david-eckardt-sociomantic: This assertion is pointless because it does the same as the array bounds checks in the function body. This condition needs to be checked with `enforce`, though, because `this.remaining` contains external input data.

gavin-norman-sociomantic: Ah yes, I wondered if someone would notice that. I included it as an in contract so that it'll be translated into a `verify` when we convert swarm. You're right, though, `enforce` also makes sense.

gavin-norman-sociomantic:
Updated: now with doc and release notes.

gavin-norman-sociomantic:
Updated: D2 const fixes in unittests.

gavin-norman-sociomantic:
Forgot one ;)

src/swarm/neo/util/Batch.d
```diff
@@ -0,0 +1,600 @@
...
+        Returns:
+            the number of bytes the specified field would consume in the batch
+
+    ***************************************************************************/
+
+    private size_t sizeOfField ( Field ) ( Field field )
+    {
+        static if ( isDynamicArrayType!(Field) )
+        {
+            static assert(rankOfArray!(Field) == 1,
```
david-eckardt-sociomantic: `static assert(!hasIndirections!(typeof(field[0])))`, otherwise an array of structs containing arrays would be accepted for `Field`.

src/swarm/neo/util/Batch.d
```diff
@@ -0,0 +1,600 @@
...
+    private void addField ( Field ) ( Field field )
+    {
+        static if ( isDynamicArrayType!(Field) )
+        {
+            static assert(rankOfArray!(Field) == 1,
+                "Only 1d arrays and value types are supported, not "
+                ~ Field.stringof);
+            auto len = field.length;
+            this.addBytes((cast(void*)&len)[0 .. size_t.sizeof]);
+            this.addBytes(field[0..$]);
```
david-eckardt-sociomantic: `field[]`?

gavin-norman-sociomantic:
Updated.

src/swarm/neo/util/Batch.d
```diff
@@ -0,0 +1,617 @@
...
+*******************************************************************************/
+
+public scope class BatchReader ( Record ... )
+{
+    import ocean.core.Enforce;
+
+    static assert(recordFieldsSupported!(Record)());
+
+    /// Batch. Set in constructor.
+    private Const!(void)[] batch;
```
david-eckardt-sociomantic: This doesn't seem to be used anywhere.

gavin-norman-sociomantic: :+1: Updated. (Also added comment to ctor arg about lifetime of buffer.)

src/swarm/neo/util/Batch.d
```diff
@@ -0,0 +1,615 @@
...
+            decompress_buf = buffer into which the batch is uncompressed. The
+                buffer must remain unmodified and accessible for the lifetime of
+                this instance
+
+    ***************************************************************************/
+
+    public this ( Lzo lzo, in void[] batch, ref void[] decompress_buf )
+    {
+        // Read uncompressed length from first size_t.sizeof bytes.
+        auto uncompressed_len = *(cast(size_t*)(batch.ptr));
```
david-eckardt-sociomantic: This needs validation since the content of `batch` comes from an external source. (Even the array bounds check is disabled here! :frowning_face:)
For the same reason `Lzo.uncompress` is [unsafe](https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/io/compress/Lzo.d#L113), use [`decompressSafe`](https://github.com/sociomantic-tsunami/ocean/blob/v3.x.x/src/ocean/io/compress/Lzo.d#L152) instead.

gavin-norman-sociomantic: Good catch.

gavin-norman-sociomantic: Updated.

src/swarm/neo/util/Batch.d
```diff
@@ -0,0 +1,619 @@
...
+
+*******************************************************************************/
+
+private bool recordFieldsSupported ( Record ... ) ( )
+{
+    foreach ( Field; Record )
+    {
+        static if ( isDynamicArrayType!(Field) )
+        {
+            static if ( rankOfArray!(Field) > 1 )
```
david-eckardt-sociomantic: This should be implied by the `hasIndirections` check.

david-eckardt-sociomantic: For the sake of an error message that is more comprehensive than "something is wrong" I suggest using `static assert`s on the error conditions with the unsupported type and maybe even the field name in the message, and naming the function void `checkRecordFieldsSupported` or something similar.

gavin-norman-sociomantic: :+1: updated.

david-eckardt-sociomantic (approved):

PR #191, opened 2017-12-08
Upgrade to ocean v4 RC 1

mihails-strasuns-sociomantic:
```
* submodules/ocean v4.0.0-alpha.0(6941094)...v4.0.0-alpha.3(186faa8) (110 commits)
  > `NotifyingQueue.push`: Don't use `copy`
  > Add alias to fix subclass overload set
  > Change queue methods to use `void[]` buffers with `const` where applicable
  > Small type fixes
  > Add missing cast in unittest
  (...)
```

There is one change that I don't know how to handle (unregisterSocker override),
please advise.

### DIFF
--- a/src/swarm/client/connection/RequestOverflow.d
+++ b/src/swarm/client/connection/RequestOverflow.d
@@ -293,11 +293,11 @@ public class DiskOverflow : IRequestOverflow
             Deserializes and re-assigns an overflowed request.
 
             Params:
-                stored = ubyte[] containing a serialized request
+                stored = void[] containing a serialized request
 
         ***********************************************************************/
 
-        public void restore ( ubyte[] stored );
+        public void restore ( void[] stored );
     }
 
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,8 +25,8 @@ env:
         - DMD=2.070.2.s* DIST=xenial F=devel
         - DMD=2.071.2.s* DIST=xenial F=production
         - DMD=2.071.2.s* DIST=xenial F=devel
-        - DMD=2.077.* DIST=xenial F=production
-        - DMD=2.077.* DIST=xenial F=devel
+        - DMD=2.078.* DIST=xenial F=production
+        - DMD=2.078.* DIST=xenial F=devel
 
 install: beaver dlang install
 
--- a/src/swarm/neo/node/ConnectionHandler.d
+++ b/src/swarm/neo/node/ConnectionHandler.d
@@ -489,6 +489,19 @@ class ConnectionHandler : IConnectionHandler
 
     /***************************************************************************
 
+        Called by `finalize` to unregister the connection socket from epoll
+        before closing it.
+
+    ***************************************************************************/
+
+    override protected void unregisterSocket ( )
+    {
+        if (this.connection.is_registered())
+            this.shared_params.epoll.unregister(this.connection);
+    }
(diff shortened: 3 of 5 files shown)


mihails-strasuns-sociomantic:
As for D2 failures - I think we should revert change that made `pop` return `const(void)[]` instead of a `void[]`. It causes lot of transitive const damage and is not necessary because queue internally is known to store mutable data. FYI @david-eckardt-sociomantic

Burgos:
I think just calling `this.connection.shutdown` (https://github.com/sociomantic-tsunami/swarm/blob/v4.x.x/src/swarm/neo/connection/ConnectionBase.d#L1005) will be enough. Pretty much, we want to unregister from the epoll anything that could reference this socket.

Burgos:
(I would confirm this with @david-eckardt-sociomantic )

gavin-norman-sociomantic:
>I think just calling this.connection.shutdown (https://github.com/sociomantic-tsunami/swarm/blob/v4.x.x/src/swarm/neo/connection/ConnectionBase.d#L1005) will be enough. Pretty much, we want to unregister from the epoll anything that could reference this socket.

`connection.shutdown` will actually shut down the socket and the send/receive loops. I'm not 100% sure whether that's really what we want to do here.

Another possibility would be to just do this: https://github.com/sociomantic-tsunami/swarm/blob/v4.x.x/src/swarm/neo/connection/ConnectionBase.d#L1127-L1130

@david-eckardt-sociomantic what do you think?

david-eckardt-sociomantic:
`unregisterSocket` should only unregister the socket from epoll in the way @gavin-norman-sociomantic suggested:
```d
if (socket.is_registered)
    epoll.unregister(socket);
```

mihails-strasuns-sociomantic:
Updated.

src/swarm/client/connection/RequestOverflow.d
```diff
@@ -293,11 +293,11 @@ public class DiskOverflow : IRequestOverflow
             Deserializes and re-assigns an overflowed request.
 
             Params:
-                stored = ubyte[] containing a serialized request
+                stored = void[] containing a serialized request
 
         ***********************************************************************/
 
-        public void restore ( ubyte[] stored );
+        public void restore ( void[] stored );
```
nemanja-boric-sociomantic, line 300: Breaking change? :police_car:

nemanja-boric-sociomantic: Ha, this is the interesting case where the swarm release notes are still relevant for the app users.

mihails-strasuns-sociomantic: Yup, this breaking change is already documented in ocean though (which is what forces it in swarm).

nemanja-boric-sociomantic (approved):

src/swarm/client/connection/RequestOverflow.d
gavin-norman-sociomantic: Isn't the change in ocean just an implementation detail, from the POV of a user of this module in swarm? i.e. there should be a migration note here.

gavin-norman-sociomantic:
D2 is unhappy.

mihails-strasuns-sociomantic:
That is D-APT package version issue we mentioned, added commit to switch to 2.078 to workaround it

PR #194, opened 2017-12-11
Adapt DelayedSuspender to require RequestEventDispatcher

### LINKED ISSUE #92: DelayedSuspender isn't safe to use with RequestEventDispatcher
[`DelayedSuspender`](https://github.com/sociomantic-tsunami/swarm/blob/v4.x.x/src/swarm/neo/util/DelayedSuspender.d) directly suspends and resumes a request fiber. This is fine for use directly with a request-on-conn fiber, but is risky with multi-fiber requests. The reason is that, in order for a fiber to be properly killed (== cleanly deallocated, if allocated on the stack), it must be registered with the `RequestEventDispatcher`, which handles the events from the request-on-conn, including any exceptions thrown. An exception is re-thrown in each fiber registered with the `RequestEventDispatcher`, ensuring all are killed.

gavin-norman-sociomantic:
Fixes #92.

(Done as an optional feature, rather than a breaking API change.)

### DIFF
--- a/relnotes/delayed-suspender.migration.md
+++ b/relnotes/delayed-suspender.migration.md
@@ -1,0 +1,18 @@
+## `DelayedSuspender` requires the request event dispatcher
+
+`swarm.neo.util.DelayedSuspender`
+
+Previously, using the `DelayedSuspender` with the request event dispatcher was
+buggy, as the suspender did not un/register the suspended fiber with the event
+dispatcher. This patch adapts the code to require the `DelayedSuspender` to be
+told about the existence of the request event dispatcher that's in use.
+
+Initialisation code like:
+```
+auto suspender = DelayedSuspender(fiber);
+```
+should be changed to:
+```
+auto suspender = DelayedSuspender(&request_event_dispatcher,
+    conn_event_dispatcher, fiber, SuspendSignalCode);
+```
--- a/src/swarm/neo/util/DelayedSuspender.d
+++ b/src/swarm/neo/util/DelayedSuspender.d
@@ -2,12 +2,13 @@
 
     Request handler helper to delay suspension of a fiber until a suitable time.
 
-    It is common to have multiple fibers in a request handler, one of which
-    wants to suspend another, at some point. However, it is not possible to
-    suspend a fiber from the outside; the suspension must be handled from inside
-    the fiber to be suspended. This helper mediates this situation by tracking
-    pending suspension requests and allowing the fiber to handle them
-    (suspending itself) at an appropriate juncture.
+    It is common to have multiple fibers in a request handler (managed by a
+    RequestEventDispatcher), one of which wants to suspend another, at some
+    point. However, it is not possible to suspend a fiber from the outside; the
+    suspension must be handled from inside the fiber to be suspended. This
+    helper mediates this situation by tracking pending suspension requests and
+    allowing the fiber to handle them, suspending itself at an appropriate
+    juncture.
 
     Copyright: Copyright (c) 2017 sociomantic labs GmbH. All rights reserved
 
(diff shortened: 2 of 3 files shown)


relnotes/delayed-suspender.feature.md
```diff
@@ -0,0 +1,18 @@
...
+told about the existence of the request event dispatcher that's in use.
+
+Initialisation code like:
+```
+auto suspender = DelayedSuspender(fiber);
+```
+should be changed to:
+```
+auto suspender = DelayedSuspender(fiber, &request_event_dispatcher,
+    SuspendSignalCode);
```
nemanja-boric-sociomantic: I'm a bit torn if we should actually make this `+breaking` ?

src/swarm/neo/util/DelayedSuspender.d
```diff
@@ -113,11 +134,36 @@ struct DelayedSuspender
...
+        Returns:
+            true if this.request_controller is non-null
+
+    ***************************************************************************/
+
+    private bool usingRequestEventDispatcher ( )
+    {
+        if ( this.request_event_dispatcher )
+        {
+            assert(this.conn);
```
nemanja-boric-sociomantic: `assert (this.conn !is null)`

src/swarm/neo/util/DelayedSuspender.d
```diff
@@ -113,11 +134,36 @@ struct DelayedSuspender
                 break;
             case Pending:
                 this.state = Suspended;
-                this.fiber.suspend(this.token);
+                if ( this.usingRequestEventDispatcher() )
+                this.request_event_dispatcher.nextEvent(this.fiber,
```
nemanja-boric-sociomantic: indentation.

src/swarm/neo/util/DelayedSuspender.d
```diff
@@ -30,6 +32,21 @@ struct DelayedSuspender
...
+    // change the API to make it required.
+    private RequestEventDispatcher* request_event_dispatcher;
+
+    /// The request-on-conn event dispatcher to dispatch signals via. Only used
+    /// if this.request_event_dispatcher is non-null.
+    private RequestOnConnBase.EventDispatcher conn;
+
+    /// The signal code to dispatch to resume a waiting fiber. Only used if
+    /// this.request_event_dispatcher is non-null.
+    private ubyte signal_code;
```
nemanja-boric-sociomantic: Is this set anywhere?

nemanja-boric-sociomantic: Ah, I see, it's set when you create instance.

relnotes/delayed-suspender.feature.md
gavin-norman-sociomantic: It's an optional change, though, in principle. (Though code that is not adapted will be buggy.)

nemanja-boric-sociomantic: Yeah, I'm scared of that _buggy_ bit. I wouldn't insist on it, though.

gavin-norman-sociomantic: The alternative would be to do this as a `+breaking` patch release. I suppose that's the strictly correct way.

nemanja-boric-sociomantic: Yes, that would be the strictly correct way to do this (this is what I had in mind with my first comment, sorry for not being clear for the patch release).

gavin-norman-sociomantic:
Updated to a breaking bugfix on v4.4.x.

nemanja-boric-sociomantic (dismissed):

relnotes/delayed-suspender.feature.md
mathias-baumann-sociomantic: Even cooler would be if you can make the change so that you _have_ to modify the code to use it or else compilation fails. No hidden-buggy behavior then.

nemanja-boric-sociomantic: Changing the order of the fields in the structure would work.

gavin-norman-sociomantic: Ah you're right. I can just change the order of the fields :+1:

gavin-norman-sociomantic: Updated.

relnotes/delayed-suspender.migration.md
```diff
@@ -0,0 +1,18 @@
...
+dispatcher. This patch adapts the code to require the `DelayedSuspender` to be
+told about the existence of the request event dispatcher that's in use.
+
+Initialisation code like:
+```
+auto suspender = DelayedSuspender(fiber);
+```
+should be changed to:
+```
+auto suspender = DelayedSuspender(fiber, &request_event_dispatcher,
```
nemanja-boric-sociomantic: `DelayedSuspender(&request_event_dispatcher, this.outer.conn, fiber, ...` (the order is now different, and it was missing `RoC.EventDispatcher` field).

gavin-norman-sociomantic: Updated.

nemanja-boric-sociomantic (approved):
LGTM again.

PR #198, opened 2017-12-13
Make the enum generated via genMachineState public

mathias-lang-sociomantic:
Before it was private, however it also exposed a public function which used it as an argument,
which led to a weird situation where it couldn't be named.
The old protection rules did not catch this, but with the new visibility rules from v2.071+,
it creates a deprecation with dmqproto, which uses it in the fakedmq derived class.

### DIFF
--- a/src/swarm/neo/util/StateMachine.d
+++ b/src/swarm/neo/util/StateMachine.d
@@ -121,7 +121,7 @@ unittest
 unittest
 {
     test!("==")(genStateMachine(["One", "Two", "Three"]),
-        "private enum State:uint {Exit,One,Two,Three}" ~
+        "public enum State:uint {Exit,One,Two,Three}" ~
         "private State state;" ~
         "public void run ( State init_state ) {" ~
             "this.state = init_state; do {" ~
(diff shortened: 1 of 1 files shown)


nemanja-boric-sociomantic:
I thought we removed/deprecated this?

nemanja-boric-sociomantic:
Ah, it could be still used in the `Consume_v2`.

gavin-norman-sociomantic:
Yeah, it's scheduled for deprecation (could do that in the next release), but I think some requests still use it.

mathias-lang-sociomantic:
It's still in v5.x.x in swarm: https://github.com/sociomantic-tsunami/swarm/blob/6eed1d0d4e3f316282d92d9efafdb0c05dcd5060/src/swarm/neo/util/StateMachine.d

And it's used by both v1 and v2 Consume: https://github.com/sociomantic-tsunami/dmqproto/blob/cc6f1c93ed62e8db902f6375ac6a339b713dd6c0/src/dmqproto/node/neo/request/Consume.d#L43-L47

But the override is only problematic for v1 AFAIK:
```
./src/fakedmq/neo/request/Consume.d(182): Deprecation: dmqproto.node.neo.request.Consume.ConsumeProtocol_v1.State is not visible from module Consume
```

mathias-lang-sociomantic:
Since my fixes are scheduled to go in v14.x.x in dmqproto, I'd be happy with another course of action (e.g. removing the v1).

gavin-norman-sociomantic (approved):

gavin-norman-sociomantic:
Let's just merge this. It's a harmless fix.

PR #206, opened 2017-12-14
Deprecate old core client ctors

gavin-norman-sociomantic:


gavin-norman-sociomantic:
I'd like some feedback from users on this PR too, just to make sure I'm not deprecating something that has a valid use case. I'll request some people to look at it.

nemanja-boric-sociomantic (approved):

gavin-norman-sociomantic:
Don said that his neo client apps only use the config ctor.

PR #208, opened 2017-12-14
Add facility to switch off automatic connection establishment in client

gavin-norman-sociomantic:
Based on #206.


gavin-norman-sociomantic:
Tested with a `Task` that constructs a DHT client in its constructor then connects in its `run` method. (Adapted https://github.com/sociomantic-tsunami/dhtnode/blob/neo/neotest/main.d#L27-L57.) Works as expected, including initialising connections to non-contactable nodes, and re-establishing dropped connections after initialisation.

gavin-norman-sociomantic:
#206 merged. Unblocked.

nemanja-boric-sociomantic (approved):

Issue #168, opened 2017-11-07
Could request version handling be moved into swarm?


gavin-norman-sociomantic:
Currently, every node-side request implementation has code to parse the version, check whether it's supported, and return a status code, if not. (e.g. https://github.com/sociomantic-tsunami/dhtnode/blob/neo/src/dhtnode/request/neo/GetAll.d#L41-L65.)

This could probably be moved into the neo connection handler, where it also parses the request code.

PR #209, opened 2017-12-14
Add RequestStarted to GlobalStatusCode enum

nemanja-boric-sociomantic:
This PR adds support for moving the initial "request supported" handshake into the swarm. The request should be adapted so they don't send the request implementation defined `Started` code, but to just send `RequestStarted` when they are ready to start handling the request. Then they can send `Error` message (which is request defined) any time to the client, which should also expect it and abort everything with the `node_error`.

This is to allow request handlers to be `nothrow`, notifying the client about the error that just happened, and which can happen any time.

Node's connection handler then can be enhanced with the nothrow enforcement, but that needs to be done in the next major.

### DIFF
--- a/integrationtest/neo/client/request/internal/GetAll.d
+++ b/integrationtest/neo/client/request/internal/GetAll.d
@@ -158,6 +158,7 @@ private scope class GetAllImpl
     import swarm.neo.connection.RequestOnConnBase;
     import swarm.neo.client.mixins.AllNodesRequestCore;
     import swarm.neo.client.mixins.SuspendableRequestCore;
+    import swarm.neo.request.Command;
     import swarm.neo.request.RequestEventDispatcher;
     import swarm.neo.client.RequestOnConn;
     import swarm.neo.util.MessageFiber;
@@ -273,34 +274,10 @@ private scope class GetAllImpl
 
     private bool handleStatusCode ( ubyte status )
     {
-        auto getall_status = cast(RequestStatusCode)status;
+        auto request_status = cast(SupportedStatus)status;
 
-        if ( GetAll.handleGlobalStatusCodes(getall_status, this.context,
-            this.conn.remote_address) )
-            return false; // Global code, e.g. request/version not supported
-
-        // GetAll-specific codes
-        with ( RequestStatusCode ) switch ( getall_status )
-        {
-            case Started:
-                // Expected "request started" code
-                return true;
-
-            case Error:
-                // The node returned an error code. Notify the user and
-                // end the request.
-                GetAll.Notification n;
-                n.node_error = RequestNodeInfo(
-                    this.context.request_id, conn.remote_address);
-                GetAll.notify(this.context.user_params, n);
-                return false;
-
-            default:
-                // Treat unknown codes as internal errors.
-                goto case Error;
-        }
-
-        assert(false);
+        return GetAll.handleSupportedCodes(request_status, this.context,
+            this.conn.remote_address);
     }
 
     /***************************************************************************
@@ -364,11 +341,12 @@ private struct Reader
         auto suspendable_control =
             &this.context.shared_working.suspendable_control;
 
-        bool finished;
+        bool finished, error;
         do
         {
             auto message = this.request_event_dispatcher.receive(this.fiber,
-                Message(MessageType.Record), Message(MessageType.End));
+                Message(MessageType.Record), Message(MessageType.End),
+                Message(MessageType.Error));
             with ( MessageType ) switch ( message.type )
             {
                 case Record:
@@ -391,6 +369,29 @@ private struct Reader
 
                 case End:
                     finished = true;
+
+                    if (!error)
+                    {
+                        // Acknowledge the End signal. The protocol guarantees that the node
+                        // will not send any further messages.
+                        this.request_event_dispatcher.send(this.fiber,
+                            ( RequestOnConn.EventDispatcher.Payload payload )
+                            {
+                                payload.addConstant(MessageType.Ack);
+                            }
+                        );
+                        this.conn.flush();
+                    }
+                    break;
+
+                case Error:
+                    finished = true;
+                    error = true;
+
+                    GetAll.Notification n;
+                    n.node_error = RequestNodeInfo(this.context.request_id,
+                            this.conn.remote_address);
+                    GetAll.notify(this.context.user_params, n);
                     break;
 
                 default:
@@ -398,16 +399,6 @@ private struct Reader
             }
         }
         while ( !finished );
-
-        // Acknowledge the End signal. The protocol guarantees that the node
-        // will not send any further messages.
-        this.request_event_dispatcher.send(this.fiber,
-            ( RequestOnConn.EventDispatcher.Payload payload )
-            {
-                payload.addConstant(MessageType.Ack);
-            }
-        );
-        this.conn.flush();
 
         // Kill the controller fiber.
         this.request_event_dispatcher.abort(this.controller.fiber);
--- a/integrationtest/neo/node/request/GetAll.d
+++ b/integrationtest/neo/node/request/GetAll.d
@@ -50,7 +50,7 @@ public void handle ( Object shared_resources, RequestOnConn connection,
             ed.send(
                 ( ed.Payload payload )
                 {
-                    payload.addConstant(GlobalStatusCode.RequestVersionNotSupported);
+                    payload.addConstant(SupportedStatus.RequestVersionNotSupported);
                 }
             );
             break;
--- a/integrationtest/neo/node/request/GetAll.d
+++ b/integrationtest/neo/node/request/GetAll.d
@@ -38,15 +38,24 @@ public void handle ( Object shared_resources, RequestOnConn connection,
     auto storage = cast(Storage)shared_resources;
     assert(storage);
 
+    auto ed = connection.event_dispatcher;
+
     switch ( cmdver )
     {
         case 0:
+            ed.send(
+                ( ed.Payload payload )
+                {
+                    payload.addConstant(SupportedStatus.RequestSupported);
+                }
+            );
+            ed.flush();
+
             scope rq = new GetAllImpl_v0;
             rq.handle(storage, connection, msg_payload);
             break;
 
         default:
-            auto ed = connection.event_dispatcher;
             ed.send(
                 ( ed.Payload payload )
                 {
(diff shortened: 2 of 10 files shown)


gavin-norman-sociomantic (commented):
A high-level comment that applies to all commits:

I'd suggest that we rename things a bit:
* `GlobalStatusCode` -> `SupportedCode`.
* `RequestStarted` -> `RequestSupported`.
* `handleRequestStatusCodes` -> `handleSupportedCodes`

The current sole function of this code is to indicate supported / not, so unless there are any ideas for it being used to carry any other messages, we may as well make it clear what it represents.

nemanja-boric-sociomantic:
Updated and rebased to v4.x.x. I'm not sure what to do about `GlobalStatusCode` enum. I feel `deprecating` it here would bring more harm than good, as it would propagate to all current requests.

nemanja-boric-sociomantic:
(and it's not an internal request detail, like `nextEvent`, but actually a part of the node/client protocol.)

src/swarm/neo/request/Command.d
```diff
@@ -61,6 +61,11 @@ public alias byte StatusCode;
 
 public enum GlobalStatusCode : StatusCode
 {
+    /// Both request code specified in the transmitted Command struct and the
+    /// associated request version are supported and the request will start
+    /// on this connection.
+    RequestSupported = -3,
```
gavin-norman-sociomantic, line 77: The commit message of "Add RequestStarted to GlobalStatusCode enum" is outdated.

src/swarm/neo/client/mixins/RequestCore.d
```diff
@@ -375,6 +375,57 @@ public template RequestCore ( RequestType request_type_, ubyte request_code,
...
+            remote_address = address of node which returned this status code
+
+        Returns:
+            false if `status` was a global error code and the user notified;
+            true if `status` is a RequestSupported code indicating that the
+            request is supported and that the request will start.
+
+    ***************************************************************************/
+
+    private static bool handleSupportedCodes ( StatusCode status,
```
gavin-norman-sociomantic: As we no longer intend to allow the set of codes to be "extended" per request type, this function should just accept `SupportedStatus` directly.

src/swarm/neo/request/Command.d
```diff
@@ -61,6 +61,11 @@ public alias byte StatusCode;
 
 public enum GlobalStatusCode : StatusCode
```
gavin-norman-sociomantic: Maybe add `alias GlobalStatusCode SupportedStatus;`.

src/swarm/neo/client/mixins/RequestCore.d
```diff
@@ -375,6 +375,57 @@ public template RequestCore ( RequestType request_type_, ubyte request_code,
...
+        Private helper function to handle notification of the global supported
+        codes, including the RequestSupported code.
+
+        Params:
+            status = code received from node
+            context = deserialized request context
+            remote_address = address of node which returned this status code
+
+        Returns:
+            false if `status` was a global error code and the user notified;
```
gavin-norman-sociomantic: "global error code" is outdated terminology.

src/swarm/neo/request/Command.d
gavin-norman-sociomantic: Oh, or would renaming the enum and adding a deprecated alias do the job? (As long as we change all usages inside swarm, we shouldn't get deprecation warnings.)

integrationtest/neo/client/request/internal/GetAll.d
```diff
@@ -275,32 +275,11 @@ private scope class GetAllImpl
     {
         auto getall_status = cast(RequestStatusCode)status;
```
gavin-norman-sociomantic: This variable is a bit deceptive now, as there are no special GetAll status codes.

integrationtest/neo/client/request/internal/GetAll.d
```diff
@@ -393,21 +373,34 @@ private struct Reader
...
-        // Acknowledge the End signal. The protocol guarantees that the node
-        // will not send any further messages.
-        this.request_event_dispatcher.send(this.fiber,
-            ( RequestOnConn.EventDispatcher.Payload payload )
-            {
-                payload.addConstant(MessageType.Ack);
-            }
-        );
-        this.conn.flush();
+        if (!error)
```
gavin-norman-sociomantic: Could this be moved inside `case End:`?

integrationtest/neo/node/request/GetAll.d
```diff
@@ -38,15 +38,23 @@ public void handle ( Object shared_resources, RequestOnConn connection,
...
+
     switch ( cmdver )
     {
         case 0:
+            ed.send(
+                ( ed.Payload payload )
+                {
+                    payload.addConstant(GlobalStatusCode.RequestSupported);
+                }
+            );
```
gavin-norman-sociomantic: This ought to be flushed.

gavin-norman-sociomantic: (All the more reason for https://github.com/sociomantic-tsunami/swarm/issues/168.)

integrationtest/neo/node/request/GetAll.d
```diff
@@ -218,33 +226,36 @@ private scope class GetAllImpl_v0
...
+                this.writer.suspender.requestSuspension();
 
-        this.controller.fiber.start();
-        this.writer.fiber.start();
-        this.request_event_dispatcher.eventLoop(this.conn);
+            this.controller.fiber.start();
+            this.writer.fiber.start();
+            this.request_event_dispatcher.eventLoop(this.conn);
+        }
+        catch (Exception e)
```
gavin-norman-sociomantic, line 253: I really wonder if this would after all be worth handling at a higher level. The only reason I can think of to not do that is that the exact code to send is defined per request (atm).

src/swarm/neo/request/Command.d
nemanja-boric-sociomantic: No, sadly, this is used outside the swarm (the node handlers are sending these codes to the clients).

gavin-norman-sociomantic: Right, that would mean that when we update *proto to use swarm v4.5.0, we'd get deprecation warnings that we'd then fix. Or?

nemanja-boric-sociomantic: Right, that would work, it's just renaming in this case.

integrationtest/neo/node/request/GetAll.d
nemanja-boric-sociomantic: It's a tricky situation. The reason I didn't want it there, is that perhaps different requests may define different ways how the errors are reported to the client (and it requires change of the request definitions to expect it anytime). I guess I could add the separate method that runs the handler, accepts the error code (or use the swarm defined, in which case we'll have to add one), but then it's the question how should we plug this with the existing request.

Perhaps we should just add the request code and send it, as, if the client is not aware of it, it will shut down with the protocol error, which is still better than the current situation?

gavin-norman-sociomantic: I suppose we can leave it down to each request for now, then see what common pattern develops. (I suspect all requests will end up doing the same, but we may be surprised.)

nemanja-boric-sociomantic:
Updated.

integrationtest/neo/client/request/internal/GetAll.d
```diff
@@ -273,7 +274,7 @@ private scope class GetAllImpl
 
     private bool handleStatusCode ( ubyte status )
     {
-        auto request_status = cast(RequestStatusCode)status;
+        auto request_status = cast(SupportedStatus)status;
```
gavin-norman-sociomantic: This looks unrelated to this commit?

nemanja-boric-sociomantic: Hm, I wanted to get rid of all the deprecations, but I should actually get rid of them in a good way - actually using the new mechanism.

nemanja-boric-sociomantic:
Updated.

gavin-norman-sociomantic:
Hmm release notes? The deprecation could be mentioned, say.

nemanja-boric-sociomantic:
Oh, damn, that slipped me!

gavin-norman-sociomantic:
Yeah, and new feature.

nemanja-boric-sociomantic:
Updated with :police_car:

src/swarm/neo/client/mixins/RequestCore.d
```diff
@@ -375,6 +375,57 @@ public template RequestCore ( RequestType request_type_, ubyte request_code,
...
+            remote_address = address of node which returned this status code
+
+        Returns:
+            false if `status` was a status error code and the user notified;
+            true if `status` is a RequestSupported code indicating that the
+            request is supported and that the request will start.
+
+    ***************************************************************************/
+
+    private static bool handleSupportedCodes ( SupportedStatus status,
```
gavin-norman-sociomantic, line 395: Actually, is there any reason to not deprecate `handleGlobalStatusCodes`?

nemanja-boric-sociomantic: Yes, we don't have suitable replacement for the old requests which can't yet move to the new handshake mechanism.

gavin-norman-sociomantic: Mmm, isn't `handleSupportedCodes` it?

nemanja-boric-sociomantic: No, because `handleSupportedCodes` can't tell the client if the actual status code was (previously) `GlobalStatusCode` or a request specific code - it returns false for all of them (it returns true only if the `RequestStarted` happens, which is never the case in the old request). In contrast, `handleGlobalStatusCodes` would return `true` if the status code was a global status code, but `false` if it was the request-specific status code.

Perhaps the deprecation path could be checking if the code is positive after casting it to `SupportedStatus` (signed byte), but that doesn't strike me as a nice solution.

gavin-norman-sociomantic: Maybe I'm missing something...

A replacement for a deprecated method doesn't have to be a _direct_ replacement. It may require changes other than changing the name of the symbol used. The deprecation just indicates that something is changing and that all users of the code will have to migrate. Which is exactly the case we have here. Or?

nemanja-boric-sociomantic: The problem is that I don't see a way to fix the old, but yet supported requests, without the change in the _protocol_ (this is in contrast to our usual deprecations that can be fixed without changing the on-wire protocol). That means that either we need to make a new version of the request, or we're stuck with the silly deprecation warning for which the proto users (so, not swarm users, but the users of the concrete proto implementations) can't do anything. Now, making a new version of the request just for this change, to get rid of this particular deprecation might be fine, but I was erring on the side of not having to do this.

nemanja-boric-sociomantic: (I probably didn't use `erring` in a good context here).

gavin-norman-sociomantic: Ah I see. You're right.

gavin-norman-sociomantic (approved):

PR #211, opened 2017-12-15
Rename test/* into integrationtest/*

nemanja-boric-sociomantic:
Makd v2.x.x expects the integration test suite inside
integrationtest/ directory, not inside test/. This renames
test/ into integrationtest/, guards `main` methods with
`version(!UnitTest)` and runs the integration tests again.

### DIFF
--- a/integrationtest/client_auth/main.d
+++ b/integrationtest/client_auth/main.d
@@ -10,9 +10,9 @@
 
 *******************************************************************************/
 
-module test.client_auth.main;
+module integrationtest.client_auth.main;
 
-import test.neo.client.Client;
+import integrationtest.neo.client.Client;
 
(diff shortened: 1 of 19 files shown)


mathias-lang-sociomantic:
This is already done in v5, what's the reason to do it in v4 ?

nemanja-boric-sociomantic:
Updated makd.

mathias-lang-sociomantic:
Oh... So the test in v4.x.x are not actually running anymore ?

nemanja-boric-sociomantic:
They haven't for a brief amount of time (yesterday). But they are all passing.

mathias-lang-sociomantic:
History repeats itself :thinking:

PR #217, opened 2017-12-21
Remove array length hack in RequestOnConnBase.Payload.addArray

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Rebased onto v4.x.x as a feature and a deprecation.

src/swarm/neo/connection/RequestOnConnBase.d
```diff
@@ -364,19 +364,19 @@ abstract class RequestOnConnBase
             /*******************************************************************
 
                 Fixed-size buffer for storing non-lvalues on the stack during
-                sending (see addConstant()).
+                sending (see addCopy()).
 
             *******************************************************************/
 
-            private ubyte[16] constants;
```
gavin-norman-sociomantic, line 371: Note this change in stack allocation. Testing with https://github.com/sociomantic-tsunami/dhtproto/pull/76 indicated that it's necessary, now this buffer is being used for array lengths and not just the odd constant.

I was wondering if this could be classified as a breaking change, but I don't think it can. Any refactoring could lead to changes in stack requirements.

david-eckardt-sociomantic (approved):

PR #229, opened 2018-01-11
Modify per-version request handler support to use ClassInfo, not handler function

### LINKED ISSUE #227: Add support for request handler scope class instantiation, based on request code/version
Traditionally, node-side request handlers have looked like this:
```D
void handler ( ... )
{
  switch ( ver )
  {
    case 0: // supported
      scope rq = new Request_v0;
      rq.handle();
      break;

    default: // unsupported
      sendUnsupportedCode();
      break;
  }
}
```
#218 moved a part of this boilerplate -- checking which versions of a request are supported by the node and sending the appropriate un/supported code -- into swarm.

The proposal is to move the rest of the boilerplate -- the instantiation of the appropriate scope class -- into swarm as well.

The impetus for this is to fix the issue discovered in https://github.com/sociomantic-tsunami/dhtproto/pull/93:
* The initial message payload of a request is a slice of the socket input buffer.
* Hence, when the request-on-conn suspends, the contents of the initial message payload may change.
* Moving the sending of the un/supported code into swarm changed the behaviour of request handlers, in that the handler might be called *after* the request-on-conn has been suspended and resumed. Thus, the contents of the initial payload buffer might have changed. = data corruption.

This proposal will fix this issue because if swarm knows about the request handler scope class, then it can call special methods of it: namely a method to which the initial payload is passed *before any RoC suspension*.

A nice side-effect of this fix is that the request handler implementation is simplified by the removal of the boilerplate.

gavin-norman-sociomantic:
The impetus for this change is to fix the issue discovered in
sociomantic-tsunami/dhtproto#93:
* The initial message payload of a request is a slice of the socket input
  buffer.
* Hence, when the request-on-conn suspends, the contents of the initial
  message payload may change.
* Moving the sending of the un/supported code into swarm changed the behaviour of
  request handlers, in that the handler might be called after the
  request-on-conn has been suspended and resumed. Thus, the contents of the
  initial payload buffer might have changed. = data corruption.

This change also has the nice side-effect of moving a lot of request
handling boilerplate into swarm.

Fixes #227.

(Note that this is a breaking change to an as yet unreleased new feature of v4.6.0.)

### DIFF
--- a/integrationtest/neo/node/Node.d
+++ b/integrationtest/neo/node/Node.d
@@ -34,10 +34,10 @@ public class Node : NodeBase!(ConnHandler)
     import integrationtest.neo.common.RequestCodes;
     import integrationtest.neo.node.Storage;
 
-    import Get = integrationtest.neo.node.request.Get;
-    import GetAll = integrationtest.neo.node.request.GetAll;
-    import Put = integrationtest.neo.node.request.Put;
-    import DoublePut = integrationtest.neo.node.request.DoublePut;
+    import integrationtest.neo.node.request.Get;
+    import integrationtest.neo.node.request.GetAll;
+    import integrationtest.neo.node.request.Put;
+    import integrationtest.neo.node.request.DoublePut;
 
     /***************************************************************************
 
@@ -65,13 +65,13 @@ public class Node : NodeBase!(ConnHandler)
         options.epoll = epoll;
 
         options.requests.add(Command(RequestCode.Get, 0),
-            "Get", &Get.handle_v0);
+            "Get", GetImpl_v0.classinfo);
         options.requests.add(Command(RequestCode.GetAll, 0),
-            "GetAll", &GetAll.handle_v0);
+            "GetAll", GetAllImpl_v0.classinfo);
         options.requests.add(Command(RequestCode.Put, 0),
-            "Put", &Put.handle_v0);
+            "Put", PutImpl_v0.classinfo);
         options.requests.add(Command(RequestCode.DoublePut, 0),
-            "DoublePut", &DoublePut.handle_v0);
+            "DoublePut", DoublePutImpl_v0.classinfo);
 
         options.credentials_map["dummy"] = Key.init;
         options.shared_resources = this.shared_resources;
@@ -95,6 +95,24 @@ public class Node : NodeBase!(ConnHandler)
     override protected cstring id ( )
     {
         return "example";
+    }
+
+    /***************************************************************************
+
+        Scope allocates a request resource acquirer backed by the protected
+        `shared_resources`. (Passed as a generic Object to avoid templatising
+        this class and others that depend on it.)
+
+        Params:
+            handle_request_dg = delegate that receives a resources acquirer and
+                initiates handling of a request
+
+    ***************************************************************************/
+
+    override protected void getResourceAcquirer (
+        void delegate ( Object resource_acquirer ) handle_request_dg )
+    {
+        handle_request_dg(this.shared_resources);
     }
 }
 
--- a/integrationtest/neo/node/Node.d
+++ b/integrationtest/neo/node/Node.d
@@ -39,9 +39,6 @@ public class Node : NodeBase!(ConnHandler)
     import Put = integrationtest.neo.node.request.Put;
     import DoublePut = integrationtest.neo.node.request.DoublePut;
 
-    /// Storage engine.
-    private Storage storage;
-
     /***************************************************************************
 
         Constructor.
@@ -60,7 +57,9 @@ public class Node : NodeBase!(ConnHandler)
     }
     body
     {
-        this.storage = new Storage;
+        // In this simple example node implementation, we don't need any shared
+        // resources except the reference to the storage.
+        this.shared_resources = new Storage;
 
         Options options;
         options.epoll = epoll;
@@ -75,7 +74,7 @@ public class Node : NodeBase!(ConnHandler)
             "DoublePut", &DoublePut.handle_v0);
 
         options.credentials_map["dummy"] = Key.init;
-        options.shared_resources = this.storage;
+        options.shared_resources = this.shared_resources;
 
         const backlog = 1_000;
         auto legacy_port = NodeItem(addr.dup, cast(ushort)(neo_port - 1));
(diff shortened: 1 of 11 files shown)


src/swarm/neo/node/ConnectionHandler.d
```diff
@@ -357,6 +375,43 @@ class ConnectionHandler : IConnectionHandler
...
+    /***************************************************************************
+
+        Re-usable buffer used to emplace request handler objects into. (Instead
+        of heap allocating an instance every time a request is handled.) We use
+        a heap buffer, rather than a fixed-size array on the stack, in order to
+        not have to guess the maximum size of a request handler object.
+
+    ***************************************************************************/
+
+    private void[] emplace_buf;
```
nemanja-boric-sociomantic: Shouldn't this be on the `RequestOnConn` level, given that the single connection handles many requests?

gavin-norman-sociomantic: Haha, probably yes.

gavin-norman-sociomantic: That'd be a big advantage of stack allocation, actually.

nemanja-boric-sociomantic: Could we plug in swarms `AcquiredBuffers` here? So that the `handleRequest` gets a buffer and recycles it?

gavin-norman-sociomantic: Hmm... In theory, yes. The code in swarm doesn't currently know anything about any kind of shared resources, though.

nemanja-boric-sociomantic: No, I mean just to have reusable buffers, nothing else.

gavin-norman-sociomantic: What would be the advantage?

nemanja-boric-sociomantic: I wrote that comment before I've heard it's easy enough to add it as a member of the `RequestOnConn` :-)

gavin-norman-sociomantic: Updated.

src/swarm/neo/node/ConnectionHandler.d
```diff
@@ -471,9 +523,26 @@ class ConnectionHandler : IConnectionHandler
...
+                    // TODO: when the old deprecated handlers are removed, this
+                    // logic can be moved into handleRequest.
+                    this.handleRequest(*rq, connection,
+                        {
+                            this.acquirer_dg(this.shared_params.shared_resources,
+                                ( Object request_resources )
+                                {
+                                    auto rq_handler =
+                                        this.emplace!(IRequestHandler)
+                                        (connection.emplace_buf, rq.class_info);
```
nemanja-boric-sociomantic: Ah, cool :+1:. I'll never be able to intuitively figure out the fact that the `connection` is actually `RequestOnConn` :-(

gavin-norman-sociomantic: Yeah, we should probably go through and make the doc / variable names consistent, at some point.

gavin-norman-sociomantic:
(Blocked until tested in dhtproto.)

gavin-norman-sociomantic:
Three requests in dhtproto converted; ok so far: https://github.com/sociomantic-tsunami/dhtproto/pull/94

src/swarm/node/model/NeoNode.d
```diff
@@ -894,6 +894,24 @@ public class NodeBase ( ConnHandler : ISwarmConnectionHandler ) : INodeBase
...
+        Params:
+            shared_resources = opaque shared resources instance owned by the
+                node. Should be cast to the node-specific shared resources
+                object
+            handle_request_dg = delegate that receives a resources acquirer and
+                initiates handling of a request
+
+    ***************************************************************************/
+
+    abstract protected void requestAcquirer ( Object shared_resources,
```
gavin-norman-sociomantic: Oof, wait, this is a breaking change.

nemanja-boric-sociomantic: I assumed we're targeting v5.x.x :-)

gavin-norman-sociomantic: Nope. v4.6.0.

gavin-norman-sociomantic: Updated to non-breaking change.

gavin-norman-sociomantic:
Updated with a few simplifications.

src/swarm/node/model/NeoNode.d
```diff
@@ -618,7 +618,7 @@ public abstract class INodeBase : INode, INodeInfo
 
 *******************************************************************************/
 
-public class NodeBase ( ConnHandler : ISwarmConnectionHandler ) : INodeBase
+public abstract class NodeBase ( ConnHandler : ISwarmConnectionHandler ) : INodeBase
```
nemanja-boric-sociomantic: `abstract`

integrationtest/neo/node/Node.d
```diff
@@ -96,6 +96,24 @@ public class Node : NodeBase!(ConnHandler)
...
+        Params:
+            handle_request_dg = delegate that receives a resources acquirer and
+                initiates handling of a request
+
+    ***************************************************************************/
+
+    override protected void getResourceAcquirer (
+        void delegate ( Object resource_acquirer ) handle_request_dg )
+    {
+        handle_request_dg(this.shared_resources);
```
nemanja-boric-sociomantic, line 115: Can't the node base class implement this?

gavin-norman-sociomantic: In the case of this dumb test node, it could. Real node implementations need to do something quite different here, though.

gavin-norman-sociomantic:
Tests pass in dhtproto, with this change. (See https://github.com/sociomantic-tsunami/dhtproto/pull/94.) Unblocked.

src/swarm/node/model/NeoNode.d
nemanja-boric-sociomantic: This is still abstract (it shouldn't be).

gavin-norman-sociomantic: Updated.

nemanja-boric-sociomantic (approved):

PR #230, opened 2018-01-12
 Adapt RequestEventDispatcher to reuse the internal arrays 

nemanja-boric-sociomantic:
RequestEventDispatcher is now adapted to accept reusable arrays used
internally, so the instance of it doesn't need to be stored in the pool.
Instead, every request can instantiate one, backed by the pooled arrays,
avoiding to maintain the separate pool of RequestEventDispatcher.

Blocked on fine v5.x.x branch.

### DIFF
--- a/integrationtest/neo/client/request/internal/GetAll.d
+++ b/integrationtest/neo/client/request/internal/GetAll.d
@@ -276,6 +276,14 @@ private scope class GetAllImpl
         Reader reader;
         Controller controller;
 
+        // Initialise the RequestEventDispatcher with the delegate that returns
+        // (reusable) void[] arrays for the internal usage. The real implementation
+        // should pass the delegate that acquires the arrays from the pool,
+        // this implementation just returns pointers to the new slices, as we
+        // don't care about GC activity here.
+        request_event_dispatcher.initialise(
+                () { auto slices = new void[][1]; return &slices[0]; });
+
         // Note: this request heap allocates two fibers each time it is handled.
         // In a real client implementation, you would want to get these fibers
         // from a pool, to avoid allocating each time.
--- a/relnotes/requestdispatcher.migration.md
+++ b/relnotes/requestdispatcher.migration.md
@@ -1,0 +1,8 @@
+## Adapt `RequestEventDispatcher` usage to new `initialise` method: don't store in a pool
+
+`swarm.neo.request.RequestEventDispatcher`
+
+It's no longer recommended to keep the reusable RequestEventDispatcher instance,
+instead it should reuse the internal arrays (obtained by the delegate passed
+to the `initialise` method). All code using it must adapt to use the `initialise`
+method in conjunction with `AcquiredResources.getVoidBuffer()` method.
--- a/src/swarm/neo/request/RequestEventDispatcher.d
+++ b/src/swarm/neo/request/RequestEventDispatcher.d
@@ -33,8 +33,10 @@
 
 module swarm.neo.request.RequestEventDispatcher;
 
+import core.stdc.string;
 import ocean.transition;
 import ocean.core.SmartUnion;
+import swarm.neo.util.VoidBufferAsArrayOf;
 
 /*******************************************************************************
 
@@ -234,6 +236,9 @@ public struct YieldedThenResumed
            be notified of successive send readiness in the order in which they
            registered to receive this notification.)
 
+    Note that it's absolutely necessary to call initialise method on it before
+    doing any work.
+
 *******************************************************************************/
 
 public struct RequestEventDispatcher
@@ -264,11 +269,11 @@ public struct RequestEventDispatcher
     /// flag means that it's never necessary to actually remove an element from
     /// this array. This is an optimisation, as array remove operations were
     /// found (by profiling) to be taking a significant number of CPU cycles.
-    private WaitingFiber[] waiting_fibers;
+    private VoidBufferAsArrayOf!(WaitingFiber) waiting_fibers;
 
     /// Buffer into which waiting_fibers are copied when performing an iteration
     /// that can modify the contents of waiting_fibers.
-    private WaitingFiber[] waiting_fibers_to_iterate;
+    private VoidBufferAsArrayOf!(WaitingFiber) waiting_fibers_to_iterate;
 
     /// Event which has fired, to be returned by nextEvent(). (This value is set
     /// by notifyWaitingFiber(), just before resuming the waiting fiber.)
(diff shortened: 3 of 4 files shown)


src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -1010,3 +1014,43 @@ public struct RequestEventDispatcher
...
+    type.
+
+    Params:
+        buffer = buffer to remove from
+        index = position in buffer from which to remove elements
+        remove_elems = number of elements to remove, defaults to one
+
+*******************************************************************************/
+private void removeShift ( T ) ( ref T buffer, size_t index,
+    size_t remove_elems = 1 )
```
nemanja-boric-sociomantic: Not sure where to place this :thinking:

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -1010,3 +1012,43 @@ public struct RequestEventDispatcher
         fiber.resume(this.token);
     }
 }
+
+/*******************************************************************************
+
+    Ocean's ocean.core.Mutation.removeShift adapted to work on AcquiredBuffersOf
```
gavin-norman-sociomantic: Do you mean adapted to work on `VoidBufferAsArrayOf`? (Maybe I'm misunderstanding.)

integrationtest/neo/client/request/internal/GetAll.d
```diff
@@ -293,6 +293,7 @@ private scope class GetAllImpl
         Reader reader;
         Controller controller;
 
+        request_event_dispatcher.initialise(() { auto slices = new void[][1]; return &slices[0]; });
```
gavin-norman-sociomantic: Wrap that line.

gavin-norman-sociomantic: Also maybe add a comment to explain what this is about. "As the test client doesn't have a proper shared resources system, we just pass a new buffer."?

Ditto for other locations where this happens.

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -823,8 +826,8 @@ public struct RequestEventDispatcher
     {
         while ( this.queued_signals.length > 0 )
         {
-            auto code = this.queued_signals[0];
-            Array.removeShift(this.queued_signals, 0);
+            auto code = this.queued_signals.array()[0];
+            .removeShift(this.queued_signals, 0);
```
gavin-norman-sociomantic: If it's just this one location, it might be better to just use `memmove` directly, rather than implementing a generic `removeShift` function.

src/swarm/neo/request/RequestEventDispatcher.d
nemanja-boric-sociomantic: Yes, typo :+1:

src/swarm/neo/request/RequestEventDispatcher.d
nemanja-boric-sociomantic: Fair enough.

nemanja-boric-sociomantic:
Updated. Note that I've tried to make this a non-breaking change without the deprecation, but I couldn't do it, since the caller is not required to call `reset()` before using the `RequestEventDispatcher` (as is the case with the swarm tests).

gavin-norman-sociomantic:
Yeah, I noticed that.

gavin-norman-sociomantic:
Rebase and unblock?

nemanja-boric-sociomantic:
Rebased & unblocked.

relnotes/requestdispatcher.migration.md
```diff
@@ -0,0 +1,8 @@
+# Don't reuse RequestEventDispatcher instances
```
gavin-norman-sociomantic: ##

gavin-norman-sociomantic: I'd say "## Adapt `RequestEventDispatcher` usage to new `initialise` method: don't store in a pool".

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -281,24 +286,25 @@ public struct RequestEventDispatcher
...
     /// running. (Signals which occur when the fiber is suspended are passed to
     /// the fiber resume method, waking it up.)
-    private ubyte[] queued_signals;
+    private VoidBufferAsArrayOf!(ubyte) queued_signals;
 
     /***************************************************************************
 
-        Resets this instance to its initial state. This method should be called
-        when retrieving an instance from a pool or free list.
+        Initialises the RequestEventDispatcher setting it's initial state.
```
gavin-norman-sociomantic: "its"

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -281,24 +286,25 @@ public struct RequestEventDispatcher
...
+
+        Params:
+            getVoidArray = void array acquirer.
```
gavin-norman-sociomantic: Maybe add "This instance will call this delegate multiple times, to acquire the buffers it needs internally. The request's acquired resources must not be relinquished during the lifetime of this instance.".

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -821,8 +826,17 @@ public struct RequestEventDispatcher
     {
         while ( this.queued_signals.length > 0 )
         {
-            auto code = this.queued_signals[0];
-            Array.removeShift(this.queued_signals, 0);
+            auto code = this.queued_signals.array()[0];
+
+            // Shift all the elements to the left by one
+            void* src = this.queued_signals.array().ptr + 1;
```
gavin-norman-sociomantic, line 836: `this.queued_signals.array().ptr + ubyte.sizeof`? (For consistency.)

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -821,8 +826,17 @@ public struct RequestEventDispatcher
...
+            void* dst = this.queued_signals.array().ptr;
+            size_t num = ubyte.sizeof * this.queued_signals.length;
```
gavin-norman-sociomantic: Hmm shouldn't this be length - 1?

relnotes/requestdispatcher.migration.md
gavin-norman-sociomantic: (Note the double ##.)

src/swarm/neo/request/RequestEventDispatcher.d
nemanja-boric-sociomantic: Puh, that's crazy. Good catch!

src/swarm/neo/request/RequestEventDispatcher.d
nemanja-boric-sociomantic: It's actually more consistent now. Imagine replacing `queued_signals` with `int` array - in that case suggested code would move the `src` pointer four elements after the first. The existing code, however, always move the pointer to one past the first element (or one past the array, if it has only one element - which is guaranteed to have by the bound checked array access in the previous line).

nemanja-boric-sociomantic:
Updated.

src/swarm/neo/request/RequestEventDispatcher.d
gavin-norman-sociomantic: Ahh yes, I was thinking of `this.queued_signals.array().ptr + 1` as already operating on a `void*`.

gavin-norman-sociomantic (approved):

PR #231, opened 2018-01-12
 Explicitly allow manual resume in EventDispatcher.nextEvent 

### LINKED ISSUE #154: Disallow resuming fiber in EventDispatcher.nextEvent if not explicitly asked
`EventDispatcher.nextEvent` accepts the flags, such as `Receive`, which will make `nextEvent` not throw an exception if the fiber is resumed with the data received - the default behavior is to allow resuming fiber for the reasons explicitly specified in the `nextEvent` call.

This behavior allows caller not to check the reason why the fiber was resumed - except in one case: the `nextEvent` will always allow explicit resume from the code with the positive resume code. This in turn makes user always to check that the resume reason is not explicit resume (with positive resume code). As this is allowed on relatively rare occasions, it seems that it might be better if the user would explicitly allow resuming the fiber (with another `NextEventFlag`).

Since this is breaking change (the behavior will change if you don't pass the flag everywhere), targeting next major.

nemanja-boric-sociomantic:
EventDispatcher.nextEvent accepts the flags, such as Receive, which will
make nextEvent not throw an exception if the fiber is resumed with the
data received - the default behavior is to allow resuming fiber for the
reasons explicitly specified in the nextEvent call.

This behavior allows caller not to check the reason why the fiber was
resumed - except in one case: the nextEvent will always allow explicit
resume from the code with the positive resume code. This in turn makes
user always to check that the resume reason is not explicit resume (with
positive resume code). As this is allowed on relatively rare occasions,
it seems that it might be better if the user would explicitly allow
resuming the fiber (with another NextEventFlag).

This changes the logic so that the explicit `NextEventFlag.Resume` is
added, so the clients that want to be resumed should use it and they can
then drop the assertion that the fiber was not resumed manually from the
places they don't want this behaviour.

Fixes #154

### DIFF
--- a/relnotes/nextevent.migration.md
+++ b/relnotes/nextevent.migration.md
@@ -1,0 +1,19 @@
+## EventDispatcher.nextEvent doesn't implicitly allow explicit resume
+
+`swarm.neo.connection.RequestOnConnBase`
+
+The EventDispatcher.nextEvent method suspends the fiber and ensures that it is
+only resumed again for an expected reason. Previously, in addition to the
+resume reasons specified by the user when calling the method, manual resumption
+of the fiber with a positive resume code was always implicitly allowed. This
+meant that user code had to manually check for this case, if it was not
+desired.
+
+Now, manual resumption of the fiber is disallowed by default, and
+EventDispatcher.nextEvent will by default treat this as a protocol error, if it
+occurs. Code that calls nextEvent should be updated as follows:
+
+   - If you wish to allow manual fiber resumption,
+     pass the new NextEventFlags.Resume flag to nextEvent.
+   - If you wish to disallow manual fiber resumption, do nothing. For tidiness,
+     you should remove any user code that checks for this occurrence.
--- a/src/swarm/neo/connection/RequestOnConnBase.d
+++ b/src/swarm/neo/connection/RequestOnConnBase.d
@@ -258,7 +258,10 @@ abstract class RequestOnConnBase
             Receive     = 1 << 0,
 
             /// Wait for an epoll event-loop cycle to occur.
-            Yield       = 1 << 1
+            Yield       = 1 << 1,
+
+            /// Wait for explicit resume with a positive resume code
+            Resume = 1 << 2,
         }
 
         /// Type of a delegate to fill in a payload for sending.
@@ -640,6 +643,7 @@ abstract class RequestOnConnBase
             bool sending = fill_payload !is null;
             bool receiving = (flags & NextEventFlags.Receive) > 0;
             bool yielding = (flags & NextEventFlags.Yield) > 0;
+            bool explicit_resume = (flags & NextEventFlags.Resume) > 0;
 
             // Set up sending, if required.
             scope payload = this.new Payload;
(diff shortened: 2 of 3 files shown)


nemanja-boric-sociomantic:
Blocked on #153

gavin-norman-sociomantic (approved):

nemanja-boric-sociomantic:
Oh, I forgot :police_car:

nemanja-boric-sociomantic:
Updated with the migration note.

gavin-norman-sociomantic:
Rebase and unblock?

nemanja-boric-sociomantic:
Rebased & unblocked.

relnotes/nextevent.migration.md
```diff
@@ -0,0 +1,13 @@
+## EventDispatcher.nextEvent doesn't implicitly allow explicit resume
+
+`swarm.neo.connection.RequestOnConnBase`
+
+`EventDispatcher.nextEvent` method used to suspend the fiber and ensure that
```
gavin-norman-sociomantic: A bit of rewording for clarity (this is an important change):

"
The `EventDispatcher.nextEvent` method suspends the fiber and ensures that it is only resumed again for an expected reason. Previously, in addition to the resume reasons specified by the user when calling the method, manual resumption of the fiber with a positive resume code was always implicitly allowed. This meant that user code had to manually check for this case, if it was not desired.

Now, manual resumption of the fiber is disallowed by default, and `EventDispatcher.nextEvent` will by default treat this as a protocol error, if it occurs. Code that calls `nextEvent` should be updated as follows:
* If you wish to allow manual fiber resumption, pass the new `NextEventFlags.Resume` flag to `nextEvent`.
* If you wish to disallow manual fiber resumption, do nothing. For tidiness, you should remove any user code that checks for this occurrence.
"

src/swarm/neo/connection/RequestOnConnBase.d
```diff
@@ -258,7 +258,10 @@ abstract class RequestOnConnBase
             Receive     = 1 << 0,
 
             /// Wait for an epoll event-loop cycle to occur.
-            Yield       = 1 << 1
+            Yield       = 1 << 1,
+
+            /// Wait for the explicit resume with the positive resume code
```
gavin-norman-sociomantic: Wait for explicit resume with a positive resume code

nemanja-boric-sociomantic:
Rebased.

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -582,6 +578,8 @@ public struct RequestEventDispatcher
             flags = flags.Receive;
             if ( this.waitingYielders() )
```
gavin-norman-sociomantic, line 583: It occurs to me that these `waitingForX` methods should probably be combined, so as not to iterate over the array multiple times. Can be another PR.

nemanja-boric-sociomantic: Good idea, I'll follow-up with the next PR in series.

nemanja-boric-sociomantic:
Updated.

gavin-norman-sociomantic (approved):

PR #237, opened 2018-01-15
Refactor/combine RED.waitingYielders()/RED.waitingForSignals() into a…

nemanja-boric-sociomantic:
… single method

This refactors/combine the
RequestEventDispatcher.waitingYielders()/waitingForSignals() into a
single method, using a single pass over the waiting fibers array to
set the fields.

### DIFF
--- a/src/swarm/neo/request/RequestEventDispatcher.d
+++ b/src/swarm/neo/request/RequestEventDispatcher.d
@@ -579,11 +579,7 @@ public struct RequestEventDispatcher
             this.dispatchQueuedSignals(conn);
 
             RequestOnConnBase.EventDispatcher.NextEventFlags flags;
-            flags = flags.Receive;
-            if ( this.waitingYielders() )
-                flags |= flags.Yield;
-            if ( this.waitingForSignals() )
-                flags |= flags.Resume;
+            flags = flags.Receive | this.nextEventYieldResumeFlags();
 
             auto event = conn.nextEvent(flags,
                 sending ? writer.event.send.get_payload : null);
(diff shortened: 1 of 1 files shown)


nemanja-boric-sociomantic:
The diff is a bit messed up, but it's a simple change - remove `waitingForSignal`/`waitingYielders.` and replace them with the single method with suboptimal name.

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -761,35 +757,39 @@ public struct RequestEventDispatcher
...
-                waiting_fiber.event.active == waiting_fiber.event.active.signal )
-                return true;
+        {
+            if ( waiting_fiber.enabled )
+            {
+                switch (waiting_fiber.event.active)
+                {
+                    case waiting_fiber.event.active.signal:
+                        flags |= flags.Resume;
+                        if (flags & ~cast(int)(flags.Resume))
```
gavin-norman-sociomantic: Hmm... wouldn't `flags.Receive` be required to also be set for this condition to be true?

nemanja-boric-sociomantic: This just does early return if the `Yield` flag was already set, so we can stop the iteration. Unless I'm missing something? I'm just comparing to non-zero of the resulting bitwise and.

So, if there was a yield set previously:

```
[resume][yield][receive]
0       1       0
```

and-ing this with `011` (`~resume`) will yield `!= 0` so we can return immediately  .

gavin-norman-sociomantic: Ahh gotcha. Would `if ( flags & flags.Yield )` be clearer and equivalent? (As we're only talking about yield / resume.)

nemanja-boric-sociomantic: Yeah, that's the way I originally wrote this, but this is the more friendly when it comes to refactoring and adding the next code (in each case you just repeat the flag you're setting).

nemanja-boric-sociomantic: Oh, no, that's wrong!

gavin-norman-sociomantic: Hmmm... it's tricky though. You only want to return if all flags _that this function deals with_ are set. (i.e. not `Receive`.)

nemanja-boric-sociomantic: Updated.

src/swarm/neo/request/RequestEventDispatcher.d
```diff
@@ -761,35 +757,42 @@ public struct RequestEventDispatcher
...
+                        break;
+                    case waiting_fiber.event.active.yield:
+                        flags |= flags.Yield;
+                        break;
+                    default:
+                        break;
+                }
 
-        return false;
+                if (flags == all_set_flags)
```
gavin-norman-sociomantic, line 790: Much nicer :+1:

gavin-norman-sociomantic (approved):

Issue #220, opened 2017-12-28
Adapt all-nodes & batch core to automatically handle supported codes


gavin-norman-sociomantic:
As some of the mentions are public fields, this is (partly) a breaking change.

nemanja-boric-sociomantic:
Fixed in #238.

PR #238, opened 2018-01-15
Adapt all-nodes / suspendable request cores to auto handle supported codes

gavin-norman-sociomantic:
Previously, this behaviour had to be defined by each request, via a policy.

Part of #220.


gavin-norman-sociomantic:
Ah :police_car: ...

gavin-norman-sociomantic:
Updated with :police_car: .

nemanja-boric-sociomantic (approved):

PR #239, opened 2018-01-15
Use Lzo.decompressSafe inside neo Batch.

### LINKED ISSUE #210: Use LZO.decompressSafe in neo.util.Batch
In the version of ocean used by swarm v4.x.x, this function is buggy.

nemanja-boric-sociomantic:
Neo's Batch class is now internally using uncompressSafe method of the Lzo
object which checks for the buffer overflow while uncompressing the data.

Fixes #210

### DIFF
--- a/src/swarm/neo/util/Batch.d
+++ b/src/swarm/neo/util/Batch.d
@@ -308,6 +308,11 @@ public scope class BatchReader ( Record ... )
                 buffer must remain unmodified and accessible for the lifetime of
                 this instance
 
+        Throws:
+            Exception if batch is too short (if it can't contain the batch
+            length), ocean.io.compress.CompressException.CompressException if
+            decompressing failed due to the decompress buffer overflow.
+
     ***************************************************************************/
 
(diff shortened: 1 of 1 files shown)


src/swarm/neo/util/Batch.d
```diff
@@ -322,7 +322,7 @@ public scope class BatchReader ( Record ... )
 
         // Decompress into this.batch.
```
gavin-norman-sociomantic: I notice this comment is wrong.

gavin-norman-sociomantic: Maybe could you also add `Throws:` to the ctor doc? (`decompressSafe` can throw.)

gavin-norman-sociomantic (commented):
Change LGTM. A couple of doc enhancement requests while we're at it...

src/swarm/neo/util/Batch.d
nemanja-boric-sociomantic: Updated.

gavin-norman-sociomantic (approved):

Issue #200, opened 2017-12-13
Convert `assert` to `verify`


gavin-norman-sociomantic:

gavin-norman-sociomantic:
There's a total of 572 asserts in `src`. Splitting them up into the following chunks seems manageable:
* `neo/client` (206) (https://github.com/sociomantic-tsunami/swarm/pull/201 reduces this number by 82)
* Everything else in `neo` (238) https://github.com/sociomantic-tsunami/swarm/pull/241
* Everything not in `neo` (128)

gavin-norman-sociomantic:
Guidelines: https://github.com/sociomantic-tsunami/sociomantic/blob/master/Code/D/assert-vs-enforce.md#guidelines

gavin-norman-sociomantic:
This can only be done in the next major, as `verify` was added in ocean v3.4.x and swarm is using v3.1.x.

gavin-norman-sociomantic:
As @david-eckardt-sociomantic wrote #201, the honour of continuing with `neo.client` goes to him.

I'll tackle the rest of `neo`.

@nemanja-boric-sociomantic could you tackle everything not in `neo`?

nemanja-boric-sociomantic:
Yes.

nemanja-boric-sociomantic:
This is done in #241 #246 and #249

PR #241, opened 2018-01-16
Convert assert -> verify in swarm.neo (excluding client)

gavin-norman-sociomantic:
Part of #200.

### DIFF
--- a/src/swarm/neo/authentication/CredentialsFile.d
+++ b/src/swarm/neo/authentication/CredentialsFile.d
@@ -43,6 +43,7 @@ import CredDef = swarm.neo.authentication.Credentials;
 import HmacDef = swarm.neo.authentication.HmacDef;
 
 import ocean.transition;
+import ocean.core.Verify;
 import ocean.io.device.File;
 import ocean.net.util.QueryParams;
 import ocean.text.convert.Formatter;
@@ -187,11 +188,11 @@ private HmacDef.Key[istring] parseContent ( istring content,
             b |= hexToByte(hex_key[i++]);
         }
 
-        assert(i == hex_key.length);
+        assert(i == hex_key.length); // enforced above
 
         /*
          * name is a cstring slice to the istring content so we don't need
-         * to idup it in order to use it as an associative array key. Assert
+         * to idup it in order to use it as an associative array key. Verify
          * the slice is correct, and cast it.
          * Note that determining that name is a slice to content via pointer
          * comparison relies on a flat memory model. In general comparing
@@ -201,7 +202,7 @@ private HmacDef.Key[istring] parseContent ( istring content,
          * has a flat memory model.
          */
         static assert(is(typeof(content) == istring));
-        assert(content.ptr <= name.ptr &&
+        verify(content.ptr <= name.ptr &&
                name.ptr + name.length <= content.ptr + content.length);
         keys[cast(istring)name] = key;
     }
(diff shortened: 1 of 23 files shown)


nemanja-boric-sociomantic:
Travis says no.

gavin-norman-sociomantic:
? Travis says nothing at all, afaics.

nemanja-boric-sociomantic:
Did you update the code in the past few minutes? I remember seeing travis fails with the assertions triggering in D2 builds in `ConnectionBase`?

gavin-norman-sociomantic:
I rebased and pushed maybe 10 mins ago.

nemanja-boric-sociomantic:
Yes, that matches the time of my comment. I think this is just a race condition.

gavin-norman-sociomantic:
I can see a D2 problem locally, anyway. Investigating.

gavin-norman-sociomantic:
Rebased.

gavin-norman-sociomantic:
Build hung at `xenial-v2: Pulling from sociomantictsunami/dlang`. Restarting.

gavin-norman-sociomantic:
Green now.

src/swarm/neo/authentication/CredentialsFile.d
```diff
@@ -187,11 +188,11 @@ private HmacDef.Key[istring] parseContent ( istring content,
             b |= hexToByte(hex_key[i++]);
         }
 
-        assert(i == hex_key.length);
+        verify(i == hex_key.length);
```
david-eckardt-sociomantic: This is already checked by `enforce` in line 163 so it can keep being `assert`.

src/swarm/neo/protocol/socket/MessageGenerator.d
```diff
@@ -264,7 +260,7 @@ struct IoVecTracker
                 start = end;
             }
 
-            assert(start == this.length);
+            verify(start == this.length);
```
david-eckardt-sociomantic, line 263: With assertions that just verify the inner logic of a function I wonder if they should be converted. The guideline says
> The built-in `assert` should *only* be used as a means to spot issues earlier in the debugging process. It should never be relied upon for error checking in a live application, as it is not present in release builds.

"A means to spot issues earlier in the debugging process" reflects my intention why I put this assertion here. It is safe to remove it in a release build.

gavin-norman-sociomantic: Verbal discussion concluded:
* Our aim must be to make release builds of critical D2 programs, in order to avoid array bounds errors being thrown and bringing down live apps.
* In that case, "internal" assertions such as this one can be left as is or converted to `verify`s. It doesn't really matter.
* Generally, the current pass of conversion is intended to be simplistic. However, if someone notes a situation like this where an assertion can be left as is, there's no harm in doing so. (I don't feel we should go out of our way to look for such cases right now, though.)

gavin-norman-sociomantic: I feel this one is a matter of taste.

[The doc](https://github.com/sociomantic-tsunami/sociomantic/blob/master/Code/D/assert-vs-enforce.md#verify) says:
>`verify` is designed to be a tool for catching programmer errors and basic sanity / safety errors

which I think applies in this case too.

src/swarm/neo/authentication/CredentialsFile.d
gavin-norman-sociomantic: Updated.

src/swarm/neo/protocol/socket/MessageGenerator.d
david-eckardt-sociomantic: Yeah, `verify` is fine here as well.

david-eckardt-sociomantic (approved):

PR #246, opened 2018-01-17
Convert assert to verify in swarm.neo.client

david-eckardt-sociomantic:
Part of #200.

### DIFF
--- a/src/swarm/neo/client/Connection.d
+++ b/src/swarm/neo/client/Connection.d
@@ -464,7 +464,7 @@ public final class Connection: ConnectionBase
 
         this.status_ = this.status_.Connecting;
         retry(this.tryConnect(), this.send_loop, this.epoll);
-        switch (this.status_)
+        final switch (this.status_)
         {
             case this.status_.Connecting:
                 this.status_ = this.status_.Connected;
@@ -480,9 +480,14 @@ public final class Connection: ConnectionBase
                  */
                 return false;
 
-            default:
+            case this.status_.Connected, this.status_.Shutdown:
+                assert(false);
+
+            version (D_Version2) {} else default:
                 assert(false);
         }
+
+        assert(false);
     }
 
     /***************************************************************************
--- a/src/swarm/neo/client/Connection.d
+++ b/src/swarm/neo/client/Connection.d
@@ -20,6 +20,8 @@ module swarm.neo.client.Connection;
 
 import swarm.neo.connection.ConnectionBase;
 
+import ocean.core.Verify;
+
 /// ditto
 public final class Connection: ConnectionBase
 {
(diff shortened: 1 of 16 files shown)


src/swarm/neo/client/Connection.d
```diff
@@ -480,9 +480,14 @@ public final class Connection: ConnectionBase
                  */
                 return false;
 
-            default:
+            case this.status_.Connected, this.status_.Shutdown:
```
gavin-norman-sociomantic, line 485: >`assert(false)` for unexpected values are converted to `verify(false)` in the following commits.

That's a nice way of handling it :+1:.

src/swarm/neo/client/requests/Stats.d
```diff
@@ -107,7 +109,7 @@ public class Stats : IRequestStats
         ulong start_time )
     {
         auto end_time = MicrosecondsClock.now_us();
-        assert(start_time <= end_time);
```
nemanja-boric-sociomantic, line 110: Happy New Year (and the clock readjustment)! This could be in the separate commit.

david-eckardt-sociomantic: Yeah... `verify` or `enforce`, who cares, really...
(Strictly speaking, this is even a bugfix, it was asserting on input data ;)

src/swarm/neo/client/RequestSet.d
```diff
@@ -349,18 +351,15 @@ public final class RequestSet: IRequestSet
...
 
         public void startOnNewConnection ( )
-        in
         {
-            assert(this.id);
-            assert(this.type == Type.multi_node);
+            verify(this.type == Type.multi_node);
 
-            auto rq = this.id in this.outer.active_requests;
+            assert(this.id);
```
gavin-norman-sociomantic: I feel it'd be a good idea to add a small comment stating where `assert`s have already been validated with `verify` or `enforce`. Otherwise, there's always the question "was this missed?".

david-eckardt-sociomantic: Actually, this is that kind of `assert` that is needed only during early debugging stages, it is much like an `invariant`. But you're right, it deserves a comment.

src/swarm/neo/client/requests/Stats.d
david-eckardt-sociomantic: I moved it to a separate commit.

david-eckardt-sociomantic:
Updated with comments for remaining `assert`s and an extra commit for `enforce`.

gavin-norman-sociomantic (approved):

nemanja-boric-sociomantic (approved):

PR #247, opened 2018-01-17
 RecordStream: Avoid reading one char at a time, if we can help it

mathias-lang-sociomantic:
- Add some tests
- Fix some ridiculously poor performances

### DIFF
--- a/integrationtest/record_batcher/main.d
+++ b/integrationtest/record_batcher/main.d
@@ -1,0 +1,104 @@
+/*******************************************************************************
+
+    Test the behavior of RecordStream
+
+    This has to be an integration test because we test the behavior of reading
+    from stdin.
+
+    Copyright:
+        Copyright (c) 2018 sociomantic labs GmbH. All rights reserved
+
+    License:
+        Boost Software License Version 1.0. See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module integrationtest.record_batcher.main;
+
+import core.sys.posix.unistd;
+
+import ocean.core.Test;
+import ocean.io.Console;
+import ocean.io.device.Device;
+import ocean.transition;
+
+import swarm.util.RecordStream;
+
+import ocean.io.stream.Buffered;
+import core.sys.posix.unistd;
+import core.sys.posix.sys.wait;
+
+
+version (UnitTest) {} else
+public int main ()
+{
+    int[2] pipes;
+    assert(pipe(pipes) == 0);
+
+    if (auto pid = fork())
+    {
+        return runSender(pid, pipes[1]);
+    }
+    else
+    {
+        dup2(pipes[0], STDIN_FILENO);
+        runListener();
+        return 0;
+    }
+}
+
+public int runSender (pid_t pid, int write_fd)
+{
+    mstring buffer;
+    // We need the destructor to be called at the end, to close the fd
+    {
+        scope conduit = new FDConduit(write_fd);
+        foreach (record; Input)
+            record.serialize(conduit, buffer);
+    }
+
+    int status;
+    auto ret = waitpid(pid, &status, 0);
+    assert(ret > 0, "waitpid returned <= 0");
+    return WIFEXITED(status) ? WEXITSTATUS(status) : -1;
+}
+
+public void runListener ()
+{
+    size_t index;
+    scope comparer = (Record r)
+        {
+            test!("==")(r.key, Input[index].key);
+            test!("==")(r.value, Input[index].value);
+            index++;
+            return index < Input.length;
+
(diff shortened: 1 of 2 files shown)


src/swarm/util/RecordStream.d
```diff
@@ -159,6 +159,16 @@ public struct Record
         splitRecord(cast(ubyte[])buf, this.key, this.value);
     }
 
+    /// Ditto
```
gavin-norman-sociomantic: Typical comment re: ditto ;)

src/swarm/util/RecordStream.d
```diff
@@ -159,6 +159,16 @@ public struct Record
...
+    private void deserialize ( Console.Input stream )
+    {
+        cstring data;
+        stream.readln(data);
```
gavin-norman-sociomantic: Would be nice to describe what the fix is exactly, in the commit message. (You mention the effect it has, but not what the fix actually is.)

integrationtest/record_batcher/main.d
```diff
@@ -0,0 +1,115 @@
...
+{
+    mstring buffer;
+    // We need the destructor to be called at the end, to close the fd
+    {
+        scope conduit = new FDConduit(write_fd);
+        foreach (record; Input)
+            record.serialize(conduit, buffer);
+
+        sleep(2);
+        // const istring msg = "Hello world!";
```
gavin-norman-sociomantic: Dead code.

integrationtest/record_batcher/main.d
```diff
@@ -0,0 +1,115 @@
...
+
+        };
+    scope stream = new StdinRecordStream(comparer);
+    auto eos = stream.process();
+    assert(!eos, "End Of Stream reached!");
+}
+
+
+private const Input = [
+    Record(null,     [ 0x52, 0x45, 0x50, 0x4F, 0x52, 0x54, 0x49, 0x4E, 0x47 ]),
```
gavin-norman-sociomantic: Maybe say
```D
// Two records without keys (values only)
...
// Two records with keys (16 hex characters)
...
```

integrationtest/record_batcher/main.d
```diff
@@ -0,0 +1,115 @@
...
+    Record(null,     [ 0x20, 0x57, 0x41, 0x53 ]),
+    // Keys are 16 hex char
+    Record([ 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42, 0x42,
+             0x42, 0x42, 0x42, 0x42, 0x42 ],
+           [ 0x48, 0x65, 0x6C, 0x6C, 0x6F, 0x20, 0x57, 0x6F, 0x72, 0x6C, 0x64 ]),
+];
+
+private const Output = [
+    Input[0],
```
gavin-norman-sociomantic: Why is the receiver not comparing the received records against the same array as the sender is sending? (Also why one less element?)

gavin-norman-sociomantic (commented):
Thanks!!

integrationtest/record_batcher/main.d
mathias-lang-sociomantic: For the story, what happened is that I originally tested ONLY corner cases and was seeing weird results.
Basically if you `serialize` a record, it doesn't check if  they key is either `null` / empty or if it's 16 chars. So if you put `Record([ 0x20 ], [ ... ])` it will gladly serialize it, and will screw up the deserialization big time. I will try to implement a fix for this in v5.x.x after this is merged.

That being said - I will remove the `Output` array.

integrationtest/record_batcher/main.d
mathias-lang-sociomantic: It's only one record with key, though

gavin-norman-sociomantic: Oh yeah. Well, `// One record with a key (16 hex characters)` then.

mathias-lang-sociomantic:
Updated

gavin-norman-sociomantic (approved):

gavin-norman-sociomantic (approved):

mihails-strasuns-sociomantic:
It looks like this has introduced a regression in swarmtools - those now fail if input stream is large enough with:

```
terminated after throwing an uncaught instance of 'ocean.core.ExceptionDefinitions.IOException' at ./submodules/ocean/src/ocean/io/device/Conduit.d:190
  toString():  BufferedInput.next :: input buffer is full
```

mathias-lang-sociomantic:
There's also https://github.com/sociomantic-tsunami/swarm/pull/265
But it doesn't look related :(

PR #249, opened 2018-01-18
Convert assert -> verify outside neo.

nemanja-boric-sociomantic:


src/swarm/client/helper/NodesConfigReader.d
```diff
@@ -191,13 +191,13 @@ static:
         # 127.0.0.3:6000";
 
         auto res = read(content);
-        assert(res.length == 2);
+        test(res.length == 2);
```
david-eckardt-sociomantic: I think the intended usage is `test!("==")(res.length, 2)`.

nemanja-boric-sociomantic: Yes, I've missed all these.

src/swarm/client/registry/NodeRegistry.d
```diff
@@ -232,7 +232,7 @@ public abstract class NodeRegistry : INodeRegistry
         {
             auto handled =
                 this.assignClientCommand(this.getClientCommandParams(params));
-            assert(handled, "unhandled client command code");
+            enforce(handled, "unhandled client command code");
```
david-eckardt-sociomantic: This looks more like `verify` as it can fail if `params` is invalid. `enforce` is meant to verify input to the program.

src/swarm/util/Hash.d
```diff
@@ -208,17 +218,22 @@ public hash_t straightToHash ( hash_t hash )
...
     Throws:
-        asserts that the string is a valid hex hash
+        InvalidHashException if the string is not a valid hex hash
 
  **************************************************************************/
 
 public hash_t straightToHash ( cstring hash )
-in
-{
-    assert(isHash(hash), "straightToHash(cstring)  - string is not a valid hash " ~ hash);
```
gavin-norman-sociomantic, line 217: Haha. Old-school.

src/swarm/util/RecordBatcher.d
```diff
@@ -503,7 +504,7 @@ public class RecordBatch : RecordBatchBase
     {
         // Read uncompressed length from first size_t.sizeof bytes.
         auto uncompressed_len = *(cast(size_t*)(compressed.ptr));
-        assert(uncompressed_len <= this.batch.dimension);
+        enforce(uncompressed_len <= this.batch.dimension);
```
gavin-norman-sociomantic, line 507: Oof.

src/swarm/client/registry/NodeRegistry.d
nemanja-boric-sociomantic: The trick is that I suspect that the client command code could arrive from the external input, following dependencies. `assignClientCommand` may not cover all possible client command codes.

nemanja-boric-sociomantic: Wait, I'm wrong, this is done on _client_ side.

nemanja-boric-sociomantic: This should indeed be a `verify`.

gavin-norman-sociomantic: Right, the command is determined by the programmer.

src/swarm/client/connection/NodeErrorTracker.d
```diff
@@ -196,7 +197,7 @@ public class NodeErrorTracker : INodeConnectionPoolErrorReporter
 
             public bool inc ( time_t now, Weights weights )
             {
-                assert(now >= this.last_time);
+                verify(now >= this.last_time);
```
david-eckardt-sociomantic: I would use `enforce` here, time values are usually program input.

nemanja-boric-sociomantic: I was on the edge of this (since it's not clear if the time input is coming from the clock, or from the ticks). I'll move to enforce.

nemanja-boric-sociomantic: (meaning I would personally not feed this method with the non-monotonic clock)

nemanja-boric-sociomantic:
Updated with the @david-eckardt-sociomantic's comments so far.

david-eckardt-sociomantic (approved):

gavin-norman-sociomantic (approved):

david-eckardt-sociomantic:
For reference: Part of #200.

PR #250, opened 2018-01-18
Add reopenable files ext support to the client stats log 

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Pinged the people who had the application-side problem related to usage of swarm's `ClientStats`.

src/swarm/util/log/ClientStats.d
```diff
@@ -146,6 +152,40 @@ public class ClientStats : StatsLog
...
+
+    public this ( IApplication app, EpollSelectDispatcher epoll,
+        istring file_name, time_t period = default_period )
+    {
+        Appender newAppender ( istring file, Appender.Layout layout )
+        {
+            auto stream = new File(file, File.WriteAppending);
+
+            if ( auto reopenable_files_ext =
+                (cast(Application)app).getExtension!(ReopenableFilesExt) )
```
mihails-strasuns-sociomantic: This is a bit strange - why take `IApplication` only to immediately cast it to`Application`? Why not require `Application` argument at once?

src/swarm/util/log/ClientStats.d
```diff
@@ -146,6 +152,40 @@ public class ClientStats : StatsLog
     }
 
 
+    /***************************************************************************
+
+        Construct the class to use the default StatsLog appender, registers it
```
gautam-kotian-sociomantic: Maybe use `Constructs` to be consistent with the tense in the rest of the sentence?

src/swarm/util/log/ClientStats.d
```diff
@@ -146,6 +152,40 @@ public class ClientStats : StatsLog
...
+        with the applications' reopenable files extension (if present), and
```
gautam-kotian-sociomantic: From swarm's point of view, there'll only be a single application, so I guess the apostrophe should be left-shifted once?

src/swarm/util/log/ClientStats.d
```diff
@@ -146,6 +152,40 @@ public class ClientStats : StatsLog
...
+        starts the update timer.
+
+        Params:
+            app = application class
```
gautam-kotian-sociomantic: Interesting that you chose to pass the application instance. Why not just the reopenable files extension?

gavin-norman-sociomantic: The idea was that it's slightly more convenient for the user this way.

src/swarm/util/log/ClientStats.d
gavin-norman-sociomantic: Yeah true. From the POV of the normal usage, either would be fine, but accepting `Application` directly simplifies this code. Will update.

gavin-norman-sociomantic: (I copied some of this code from another location.)

gavin-norman-sociomantic:
Updated.

integrationtest/client_stats/main.d
```diff
@@ -0,0 +1,71 @@
...
+    scope (exit) sandbox.remove();
+
+    // Test that constructing a ClientStats with the ctor that accepts an
+    // IApplication successfully registers the log with the application's
+    // reopenable files extension.
+    auto app = new App;
+    auto client_stats = new ClientStats(app, new EpollSelectDispatcher,
+        "client_stats.log");
+    test!("==")(app.rof_ext.files.length, 1);
+    test!("==")(app.rof_ext.files[0].toString(), "client_stats.log");
```
nemanja-boric-sociomantic: Instead of this you could just do `test!("==")(app.rof_ext.reopenFile("client_stats.log"), true)` and then you can drop your `TestReopenableFilesExt`: https://github.com/sociomantic-tsunami/ocean/blob/d103e885bc2c77370d00311a2560f4be4c9d983b/src/ocean/util/app/ext/ReopenableFilesExt.d#L201

gavin-norman-sociomantic: Ah yeah. Good idea. Only works with ocean v3.6.0+ (swarm v4.x.x is on v3.1.0), but I've updated with multi-versioned code (which can be removed in swarm v5.x.x ;)

nemanja-boric-sociomantic: https://github.com/sociomantic-tsunami/swarm/issues/252

nemanja-boric-sociomantic (approved):

PR #258, opened 2018-01-19
Finish neo todos

### LINKED ISSUE #72: Fix TODOs in ClientSocket, RetryTimer, ConnectionBase, ConnectProtocol

david-eckardt-sociomantic:
Fixes #72.

### DIFF
--- a/src/swarm/neo/client/ClientSocket.d
+++ b/src/swarm/neo/client/ClientSocket.d
@@ -24,6 +24,8 @@ public class ClientSocket
     import core.sys.posix.netinet.in_: sockaddr_in;
     import core.stdc.errno: errno;
 
+    import ocean.util.log.Logger;
+
     import ocean.core.Enforce;
 
     debug ( ISelectClient ) import ocean.io.Stdout : Stderr;
@@ -94,11 +96,12 @@ public class ClientSocket
         if (this.socket_.fd >= 0)
         {
             /*
-             * TODO: Log a warning if this happens.
              * The socket should not exist or there is a bug. However, this
              * situation is too critical to assert or ignore it and leak file
              * descriptors.
              */
+            Log.lookup(this.classinfo.name).warn("connect: The socket " ~
+                "already exists.");
             this.socket.close();
         }
 
--- a/src/swarm/neo/client/Connection.d
+++ b/src/swarm/neo/client/Connection.d
@@ -462,8 +462,20 @@ public final class Connection: ConnectionBase
         if (!this.restart_after_shutdown)
             return false;
 
-        this.status_ = this.status_.Connecting;
-        retry(this.tryConnect(), this.send_loop, this.epoll);
+        try
+        {
+            this.status_ = this.status_.Connecting;
+            retry(this.tryConnect(), this.send_loop, this.epoll);
+        }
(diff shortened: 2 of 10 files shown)


src/swarm/neo/client/Connection.d
```diff
@@ -462,8 +462,20 @@ public final class Connection: ConnectionBase
...
+        try
+        {
+            this.status_ = this.status_.Connecting;
+            retry(this.tryConnect(), this.send_loop, this.epoll);
+        }
+        catch (TimerException e)
+        {
+            // The timer failed - This can happen only due to OS resource
+            // exhaustion.
+            Log.lookup(this.classinfo.name).error("connect: The retry " ~
```
nemanja-boric-sociomantic: It's a bit of a problem doing the log lookup in this case, since that's likely to fail. I suggest having a static module logger instead.

nemanja-boric-sociomantic: (to be more precise - it is likely to fail, since it will try open a new file, if not already open).

src/swarm/neo/client/ClientSocket.d
```diff
@@ -94,11 +96,12 @@ public class ClientSocket
         if (this.socket_.fd >= 0)
         {
             /*
-             * TODO: Log a warning if this happens.
              * The socket should not exist or there is a bug. However, this
              * situation is too critical to assert or ignore it and leak file
              * descriptors.
              */
+            Log.lookup(this.classinfo.name).warn("connect: The socket " ~
```
gavin-norman-sociomantic, line 103: Huh, that's a much nicer approach than the typical module logger. I wonder why we don't do this more commonly?

src/swarm/neo/client/RetryTimer.d
```diff
@@ -39,6 +42,10 @@ import ocean.transition;
     to expire after ~10 ms. When this timer expires it is set again, doubling
     the time delay up to 2.5 s, then keeping it constant.
 
+    Throws:
+        TimerFD.TimerException if the timer failed. This can happen only due to
```
nemanja-boric-sociomantic: `TimerException if the timer failed.`

src/swarm/neo/client/Connection.d
```diff
@@ -462,8 +462,20 @@ public final class Connection: ConnectionBase
...
             return false;
 
-        this.status_ = this.status_.Connecting;
-        retry(this.tryConnect(), this.send_loop, this.epoll);
+        try
+        {
+            this.status_ = this.status_.Connecting;
+            retry(this.tryConnect(), this.send_loop, this.epoll);
+        }
+        catch (TimerException e)
```
gavin-norman-sociomantic: What was the reason for changing this to only handle that one specific type of exception? My feeling is that we should handle all exceptions here, as we don't want the connection to get into an invalid state.

david-eckardt-sociomantic: Ah yes, I had thought only about moving the `catch` from `retry` here. But catching everything is better.

david-eckardt-sociomantic:
Updated.

gavin-norman-sociomantic (approved):

PR #262, opened 2018-01-23
Fix compile errors in legacy client plugins

gavin-norman-sociomantic:


mihails-strasuns-sociomantic:
Red

gavin-norman-sociomantic:
Very weird. Works with dmd 2.070.2 (in Travis) and 2.074.0 (locally), but fails with 2.071.2 and 2.078.*.

gavin-norman-sociomantic:
Pushed probable fix, anyway.

integrationtest/legacy_client/main.d
```diff
@@ -0,0 +1,42 @@
...
+
+    override protected void scopeRequestParams (
+        void delegate ( IRequestParams params ) dg ) { }
+}
+
+version (UnitTest){}
+else
+void main ( )
+{
+}
```
mathias-lang-sociomantic: Why can't this be a unittest ?

gavin-norman-sociomantic:
Green now.

integrationtest/legacy_client/main.d
gavin-norman-sociomantic: It felt nicer to put this in a separate module, as it's testing bringing together code from several other modules (the client base class and the plugins). It would have felt slightly odd to add such a test in any of those other modules. (Though I don't mind moving it, if you absolutely object.)

mathias-lang-sociomantic: I was missing the fact that it's bringing multiple modules together. Makes perfect sense then!

mihails-strasuns-sociomantic: Actually, I would suggest making it a unittest in a separate module with `_test.d` suffix. I think this is more fitting for "lot of modules but no environment/state" kind of tests.

gavin-norman-sociomantic: Right, that just occurred to me too.

gavin-norman-sociomantic: Updated.

mihails-strasuns-sociomantic (approved):

PR #265, opened 2018-01-25
 RecordStream: Correctly terminate on EOF

mathias-lang-sociomantic:
```
Before this change, RecordStream would read an empty record before terminating,
because there was no way for `deserialize` to give feedback.
Now it correctly terminate without attempting to process an empty record.
In order to test it, we needed to close the non-used ends of the `pipe` in the test.
```

### DIFF
--- a/src/swarm/util/RecordStream.d
+++ b/src/swarm/util/RecordStream.d
@@ -173,18 +173,23 @@ public struct Record
         Params:
             stream = input stream to read from
 
+        Returns:
+            `true` if there was data in the stream, `false` otherwise.
+
         Throws:
             EofException upon encountering the end of the input stream
 
     ***************************************************************************/
 
-    private void deserialize ( Console.Input stream )
+    private bool deserialize ( Console.Input stream )
     {
         cstring data;
-        stream.readln(data);
+        if (!stream.readln(data))
+            return false;
 
         // Extract the key and value from the line
         splitRecord(cast(Const!(ubyte)[])data, this.key, this.value);
+        return true;
     }
 
 
@@ -539,13 +544,8 @@ public class StdinRecordStream : ISuspendable
     {
         while (true)
         {
-            try
+            if (!this.record.deserialize(Cin))
             {
-                this.record.deserialize(Cin);
-            }
-            catch ( EofException e )
-            {
-                // An I/O exception (EOF) is expected when reading a key
                 this.end_of_stream = true;
                 break;
             }
--- a/integrationtest/record_stream/main.d
+++ b/integrationtest/record_stream/main.d
@@ -13,7 +13,7 @@
 
 *******************************************************************************/
 
-module integrationtest.record_batcher.main;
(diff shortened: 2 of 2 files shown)


src/swarm/util/RecordStream.d
```diff
@@ -178,13 +178,15 @@ public struct Record
 
     ***************************************************************************/
 
-    private void deserialize ( Console.Input stream )
```
gavin-norman-sociomantic, line 181: `Returns:` doc.

gavin-norman-sociomantic: By the way, is there a reason why this `deserialize` overload is private but the other one is public? (Not sure if I noticed that in the previous PR.)

gavin-norman-sociomantic (commented):
I don't really get why this fix works. Perhaps documenting the return value of `deserialize` will help. I'll let you know.

src/swarm/util/RecordStream.d
mathias-lang-sociomantic: Yeah, because it's a rather ugly hack. I basically say "We only accept Tango's class stdin", while a more carefully designed code would accept any `BufferedInput`.

mathias-lang-sociomantic: Maybe [this](https://github.com/sociomantic-tsunami/ocean/blob/6d387eaf835d85c88dbc6a6b1b1090a0863a26b5/src/ocean/io/Console.d#L105) would help

gavin-norman-sociomantic: Hmm... it appears that the idea was that `deserialize` provides feedback about EOF by throwing. (See the `catch` in `fiberMethod`.)

mathias-lang-sociomantic: Yes, that's correct. However I don't know how to cleanly do it with an exception. Should we have one exception per record ? Should `fiberMethod` provide the exception ? Also a `throw` seems rather wasteful compared to a `return false;`. So I agree that this mix of semantic is sub-ideal, but IMO it's a temporary situation.

mathias-lang-sociomantic:
> I don't really get why this fix works.

The other method throws on EOF. This ones return `false`.
Since the `cstring` is not populated when the call to `readln` returns `false`, `splitRecord` didn't do anything and we ended up passing an empty record to the output dg before this fix, which tripped some of our tools.

gavin-norman-sociomantic:
>The other method throws on EOF. This ones return false.

Ok. Is there any need to leave the exception handling code in `fiberMethod` then?

(By the way, I'm not disputing this change, just trying to understand what's going on so I can suggest a better way of documenting it.)

mathias-lang-sociomantic:
Just checked, you are right, only `StreamSerializer` throws `EofException`.

mathias-lang-sociomantic:
Ping ?

mihails-strasuns-sociomantic (approved):

PR #278, opened 2018-02-05
Update support guarantees to match ocean

gavin-norman-sociomantic:
The old support guarantees were appropriate to the quick
development pace that was happening with swarm neo. This
development is no longer so rapid and volatile, so putting
more typical support guarantees back in place now makes
sense.


gavin-norman-sociomantic:
Auto-merging. Trivial.

PR #281, opened 2018-02-06
Remove restriction on batch size in client-side `RecordBatch`

gavin-norman-sociomantic:


src/swarm/util/RecordBatcher.d
```diff
@@ -590,3 +590,60 @@ public class RecordBatch : RecordBatchBase
...
+    auto reader = new RecordBatch(lzo, 200);
+
+    mstring key, value_S, value_M, value_L;
+    key.length = 16;
+    value_S.length = 50;
+    value_M.length = 150;
+    value_L.length = 200;
+
+    // Test RecordBatcher.batchedSize()
+    test!("==")(writer.batchedSize(key, value_S), 16 + 8 + 50 + 8);
```
david-eckardt-sociomantic: The numbers are probably `key.length.sizeof + key.length + value_S.length.sizeof + value_S.length`, maybe make it explicit.

src/swarm/util/RecordBatcher.d
```diff
@@ -647,3 +647,31 @@ unittest
...
+    {
+        auto res = writer.add(format("0x{:16}", added),
+            format("{}{}{}", added, added, added));
+        if ( res == res.BatchFull )
+            break;
+        added++;
+    }
+
+    writer.compress(compressed);
+    assert(compressed.length > RecordBatcher.DefaultMaxBatchSize);
```
david-eckardt-sociomantic, line 710: This checks the length of the compressed data, but wasn't the restriction on the uncompressed batch? (If the compressed data are longer then the uncompressed will usually be longer, too, just checking if there is a misconception.)

src/swarm/util/RecordBatcher.d
gavin-norman-sociomantic: Yeah true.

src/swarm/util/RecordBatcher.d
gavin-norman-sociomantic: This line confirms that the compressed batch is too big to fit in the default initialised size of the batch reader, which is the whole point of this test. If the condition is not true, the test is meaningless.

src/swarm/util/RecordBatcher.d
gavin-norman-sociomantic: Updated.

PR #282, opened 2018-02-06
Update LibFeatures.d for v5.1.x

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Auto-merging. Trivial.

PR #283, opened 2018-02-06
Update LibFeatures.d for v4.7.x

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Auto-merging. Trivial.

PR #285, opened 2018-02-07
Add missing import + `Controllers` template test

david-eckardt-sociomantic:
DMD v2.078 reported a missing import for `verify` when compiling dmqproto.

### DIFF
--- a/src/swarm/neo/client/mixins/Controllers.d
+++ b/src/swarm/neo/client/mixins/Controllers.d
@@ -168,6 +168,8 @@ template Controllers ( )
     public class Suspendable ( ControllerInterface ) :
         Controller!(ControllerInterface), ISuspendable
     {
+        import ocean.core.Verify;
+
         /***********************************************************************
 
             Enum of possible pending state-changes.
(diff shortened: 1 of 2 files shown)


src/swarm/neo/client/mixins/Controllers.d
```diff
@@ -323,3 +325,21 @@ template Controllers ( )
         }
     }
 }
+
+/// Unit test instantiating the templates to detect semantic errors
+version (UnitTest) private class ControllersTemplateTest
```
gavin-norman-sociomantic, line 330: Any idea why the test client didn't catch this?: https://github.com/sociomantic-tsunami/swarm/blob/v4.x.x/integrationtest/neo/client/Client.d#L264-L270

david-eckardt-sociomantic: That code seems to use [`SuspendableController`](https://github.com/sociomantic-tsunami/swarm/blob/v4.x.x/src/swarm/neo/client/mixins/SuspendableRequestCore.d#L513).

gavin-norman-sociomantic: That's the internal controller implementation inside a client request handler. It may use that, but it's a different thing.

gavin-norman-sociomantic: I just looked into this a bit more. I was misunderstanding. The module that you've added the test for is intended for use in application code. The test client that I was referring to doesn't use this.

gavin-norman-sociomantic:
Internal issue discussing a fix for this (a new legacy channel mirror): https://github.com/sociomantic/core/issues/45.

gavin-norman-sociomantic:
An interesting question, though, is why only certain apps got locked up like this.

gavin-norman-sociomantic (approved):

PR #287, opened 2018-02-12
Add `swarm.util.Verify`

### LINKED ISSUE #286: Try adding verify to swarm (copy from ocean)
This will enable us to use `verify` in v4.x.x already, rather than having to continue using `assert` and risk forgetting to convert them into `verify`s when merging into v5.x.x.

david-eckardt-sociomantic:
Fixes #286.

Helper module to use `verify` from ocean v3.4 instead of `assert`, even if the ocean submodule is at a version that does not include `ocean.core.Verify`.

### DIFF
--- a/src/swarm/util/Verify.d
+++ b/src/swarm/util/Verify.d
@@ -1,0 +1,82 @@
+/*******************************************************************************
+
+    Helper module to use `verify` from ocean v3.4 instead of `assert`, even if
+    the ocean submodule is at a version that does not include
+    `ocean.core.Verify`. This is to allow for writing new swarm code that can be
+    safely merged into v5.
+
+    Copyright: Copyright (c) 2018 sociomantic labs GmbH. All rights reserved
+
+    License:
+        Boost Software License Version 1.0. See LICENSE_BOOST.txt for details.
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module swarm.util.Verify;
+
+import ocean.core.VersionCheck;
+
+static if (hasFeaturesFrom!("ocean", 3, 4))
+    public import ocean.core.Verify;
+else:
+
+import ocean.transition;
+
+/*******************************************************************************
+
+    Verifies that certain condition is met.
+
+    Params:
+        ok = boolean condition to check
+        msg = optional exception message
+
+    Throws:
+        SanityException if `ok` condition is `false`.
+
+*******************************************************************************/
+
+public void verify ( bool ok, lazy istring msg = "",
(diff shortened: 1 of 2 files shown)


src/swarm/util/Verify.d
```diff
@@ -0,0 +1,81 @@
+/*******************************************************************************
+
+    Helper module to use `verify` from ocean v3.4 instead of `assert`, even if
+    the ocean submodule is at a version that does not include
+    `ocean.core.Verify`. This is to allow for writing new swarm code that can be
+    safely merged into v5.
+
+    Copyright: Copyright (c) 2017 sociomantic labs GmbH. All rights reserved
```
nemanja-boric-sociomantic: 2018

nemanja-boric-sociomantic (approved):

src/swarm/util/Verify.d
```diff
@@ -0,0 +1,81 @@
...
+        Alternatively, this file may be distributed under the terms of the Tango
+        3-Clause BSD License (see LICENSE_BSD.txt for details).
+
+*******************************************************************************/
+
+module swarm.util.Verify;
+
+import ocean.core.VersionCheck;
+
+static if (hasFeaturesFrom!("ocean", 3, 4)) public import ocean.core.Verify;
```
mihails-strasuns-sociomantic: Please, at least split the statement onto own line ;)

```D
static if (hasFeaturesFrom!("ocean", 3, 4))
    public import ocean.core.Verify;
else:
```

mihails-strasuns-sociomantic (changes requested):

mihails-strasuns-sociomantic:
I think this module can be deprecated in v5 branch right after being merged into it, correct?

src/swarm/util/Verify.d
david-eckardt-sociomantic: ok, ok

nemanja-boric-sociomantic:
I wonder if we can deprecate it right away :thinking: . Edit: Ah, no we can't/there's no point anyway.

src/swarm/util/Verify.d
mihails-strasuns-sociomantic: <3

mihails-strasuns-sociomantic:
That would be rather noisy for users of old swarm

src/swarm/util/Verify.d
david-eckardt-sociomantic: It is not a statement but an import declaration :P

david-eckardt-sociomantic:
> I think this module can be deprecated in v5 branch right after being merged into it, correct?

Yes, it is for v4 only.

david-eckardt-sociomantic:
Updated with a proper line break.

mihails-strasuns-sociomantic (approved):

PR #291, opened 2018-02-23
Deprecate TCP_CORK based flush.

nemanja-boric-sociomantic:
The existing `flush` method relied on the TCP_CORK being set and it would then
pull out and put the cork back in.  However this doesn't work, because putting
the cork back in had to be done after all the packets are actually sent,
otherwise the last incomplete packet will be delayed for the 200ms. Since we
moved to the explicit application buffering for the large data and to the
explicit flushing for the control messages this flush was deprecated.

See sociomantic-tsunami/dmqproto#48

### DIFF
--- a/integrationtest/neo/node/Node.d
+++ b/integrationtest/neo/node/Node.d
@@ -64,6 +64,7 @@ public class Node : NodeBase!(ConnHandler)
 
         Options options;
         options.epoll = epoll;
+        options.no_delay = true;
 
         options.requests.add(Command(RequestCode.Get, 0),
             "Get", GetImpl_v0.classinfo);
--- a/src/swarm/neo/connection/ConnectionBase.d
+++ b/src/swarm/neo/connection/ConnectionBase.d
@@ -125,6 +125,7 @@ abstract class ConnectionBase: ISelectClient
 
         ***********************************************************************/
 
+        deprecated("Use TCP_NODELAY and explicit buffering instead.")
         public bool flush_requested;
 
         /***********************************************************************
@@ -433,8 +434,6 @@ abstract class ConnectionBase: ISelectClient
                 // Register the socket for both input and output
                 this.outer.registerEpoll(true);
                 this.outer.sender.finishSending(this.suspend());
-                if (this.flush_requested)
-                    this.outer.sender.flush();
             }
         }
     }
@@ -1208,6 +1207,7 @@ abstract class ConnectionBase: ISelectClient
 
     ***************************************************************************/
 
+    deprecated("Use TCP_NODELAY and explicit buffering instead.")
     public void flush ( )
     {
         // Flush the OS-maintained (TCP_CORK) socket output buffer. This will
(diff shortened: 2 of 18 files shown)


src/swarm/neo/protocol/socket/MessageSender.d
```diff
@@ -5,6 +5,18 @@
     This class is suitable for both half and full duplex because it does not
     handle epoll notifications directly (it is not a select client).
 
+    Once upon the time there was an explicit flush method. It relied on the
```
mathias-lang-sociomantic: I believe the right wording is "Once upon a time" :P

nemanja-boric-sociomantic: I'm just happy I put any article at all ;-)

nemanja-boric-sociomantic: Fixed, thanks!

src/swarm/neo/connection/ConnectionBase.d
```diff
@@ -433,8 +433,6 @@ abstract class ConnectionBase: ISelectClient
                 // Register the socket for both input and output
                 this.outer.registerEpoll(true);
                 this.outer.sender.finishSending(this.suspend());
-                if (this.flush_requested)
```
gavin-norman-sociomantic, line 436: Should `this.flush_requested` be deprecated as well?

integrationtest/neo/node/Node.d
```diff
@@ -64,6 +64,7 @@ public class Node : NodeBase!(ConnHandler)
 
         Options options;
         options.epoll = epoll;
+        options.no_delay = true;
```
gavin-norman-sociomantic, line 67: Is no-delay already set on the client side?

gavin-norman-sociomantic:
I do wonder if now would be a good time to start some kind of README about neo from the internal developer point-of-view. The READMEs we currently have describe the library from the user's POV, but I think it'd also be worth having a central place to put notes on using the library to implement protocols, and various design choices / experience.

src/swarm/neo/connection/ConnectionBase.d
nemanja-boric-sociomantic: Yes, good catch.

integrationtest/neo/node/Node.d
nemanja-boric-sociomantic: It is: https://github.com/sociomantic-tsunami/swarm/blob/v4.x.x/integrationtest/neo/client/Client.d#L635 (`enableSocketNoDelay()`)

nemanja-boric-sociomantic:
I think the swarm's documentation should be from the internal developer point-of-view. I see little reason that user should read swarm documentation, since the only way they are interfacing with swarm is via protos.

nemanja-boric-sociomantic:
Updated with deprecation of the `flush_requested`.

gavin-norman-sociomantic (approved):

PR #293, opened 2018-03-01
Change `EventDispatcher.shutdownWithProtocolError` to throw, not return

gavin-norman-sociomantic:
The returned exception is intended to always be thrown, and it would be a
bug, really, if the caller did not do so. Therefore, making the method
throw directly simplifies usage and removes one vector for bugs.

### DIFF
--- a/src/swarm/neo/connection/RequestOnConnBase.d
+++ b/src/swarm/neo/connection/RequestOnConnBase.d
@@ -511,24 +511,23 @@ abstract class RequestOnConnBase
             Initiates a connection shutdown from inside the fiber, throwing an
             exception of type ProtocolError in all other requests that are
             currently using this connection or attempt to use it, until the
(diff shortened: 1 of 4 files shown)


src/swarm/neo/connection/RequestOnConnBase.d
```diff
@@ -528,7 +527,7 @@ abstract class RequestOnConnBase
         {
             auto e = this.outer.connection.protocol_error.set(msg, file, line);
             this.shutdownConnection(e);
-            return e;
+            throw e;
```
nemanja-boric-sociomantic, line 530: `void shutdownWithProtocolError`

gavin-norman-sociomantic: Haha yeah. Updated.

nemanja-boric-sociomantic (approved):

PR #308, opened 2018-03-12
Remove unused import for core.memory that was causing gc_usage clash

nemanja-boric-sociomantic:
gc_usage is defined in both ocean.transition and core.memory, and we're
using the former. This commit removes import to core.memory to prevent
clashes on the symbol with the same name.


nemanja-boric-sociomantic:
Confirmed that dhtnode compiles with this.

gavin-norman-sociomantic (approved):

PR #310, opened 2018-03-14
 Throw on received message if the RoC of an all-nodes request has finished

### LINKED ISSUE #309: Terminated RequestOnConn fiber resumed on unexpected message
On a protocol mismatch the following situation can happen for an all-nodes request:
- Multiple `RequestOnConn` handler fibers are started, one for each node, and connect to the nodes.
- One of the fibers receives a message, fails to interpret correctly and terminates without shutting the connection down. (The fact that it does not initiate a shutdown is another issue; in any case the following should not happen.)
- When terminating, the fiber calls `RequestOnConnSet.finished`, which leaves the `RequestOnConn` object registered with the `RequestOnConnSet`. The request is still active because the other fibers haven't started handling the request yet.
- The node for which the client's `RequestOnConn` handler fiber has just terminated sends another message for the request.
- `Connection.setReceivedPayload` looks up the request by ID, from that it gets the -- terminated but still registered -- `RequestOnConn` handler by node address, and calls `setReceivedPayload`.
- `setReceivedPayload` attempts to resume the terminated `RequestOnConn` fiber. :boom:

The obvious solution would be to make `RequestOnConnSet.finished` remove the `RequestOnConn` object from the set. Then `Connection.setReceivedPayload` would not be able to find it. (It would then silently ignore the received message; if this is a good thing to do here may be subject to discussion.)

This is the original stack trace:
```
core.exception.AssertError@./submodules/swarm/src/swarm/neo/util/MessageFiber.d(559): attempt to resume a non-held fiber                                                                                                                                                                                                      
src/core/exception.d:438 _d_assert_msg
./submodules/swarm/src/swarm/neo/util/MessageFiber.d:559 _D5swarm3neo4util12MessageFiber17OceanMessageFiber6resumeMFS5swarm3neo4util12MessageFiber17OceanMessageFiber5TokenC6ObjectS5ocean4core10SmartUnion75__T10SmartUnionTS5swarm3neo4util12MessageFiber17OceanMessageFiber8Message_Z10SmartUnionZ9__requireMFZv
./submodules/swarm/src/swarm/neo/util/MessageFiber.d:558 ocean.core.SmartUnion.SmartUnion!(swarm.neo.util.MessageFiber.OceanMessageFiber.Message_).SmartUnion swarm.neo.util.MessageFiber.OceanMessageFiber.resume(swarm.neo.util.MessageFiber.OceanMessageFiber.Token, Object, ocean.core.SmartUnion.SmartUnion!(swarm.neo.util.MessageFiber.OceanMessageFiber.Message_).SmartUnion)./submodules/swarm/src/swarm/neo/connection/RequestOnConnBase.d:1734 void swarm.neo.connection.RequestOnConnBase.RequestOnConnBase.resumeFiber_(ocean.core.SmartUnion.SmartUnion!(swarm.neo.util.MessageFiber.OceanMessageFiber.Message_).SmartUnion)
./submodules/swarm/src/swarm/neo/connection/RequestOnConnBase.d:1779 void swarm.neo.connection.RequestOnConnBase.RequestOnConnBase.setReceivedPayload(const(void)[])
./submodules/swarm/src/swarm/neo/client/RequestOnConn.d:603 void swarm.neo.client.RequestOnConn.RequestOnConn.setReceivedPayload(const(void)[])
./submodules/swarm/src/swarm/neo/client/Connection.d:639 void swarm.neo.client.Connection.Connection.setReceivedPayload(ulong, const(void)[])
./submodules/swarm/src/swarm/neo/connection/ConnectionBase.d:651 void swarm.neo.connection.ConnectionBase.ConnectionBase.ReceiveLoop.receivedMessage(swarm.neo.protocol.Message.MessageType, const(void)[])
./submodules/swarm/src/swarm/neo/protocol/socket/MessageReceiver.d:214 void swarm.neo.protocol.socket.MessageReceiver.MessageReceiverBase.receive(lazy ocean.sys.Epoll.epoll_event_t.Event, scope void delegate(swarm.neo.protocol.Message.MessageType, const(void)[]), bool)
./submodules/swarm/src/swarm/neo/connection/ConnectionBase.d:603 void swarm.neo.connection.ConnectionBase.ConnectionBase.ReceiveLoop.fiberMethod()
src/core/thread.d:4277 void core.thread.Fiber.run()
src/core/thread.d:3523 fiber_entryPoint
```
Although we reproduced the same stack trace on a protocol mismatch and determined the aforementioned chain of events, it originally happened when a client using swarm v4.4.0 was connected to two nodes with a compatible protocol, and one of the nodes was restarted. We still need to figure out how this happens on a node restart.
We were unable to reproduce this behaviour with swarm v5.0.3.

david-eckardt-sociomantic:
In several situations, including a node shutdown and client/node protocol mismatch, the client can receive a message for a request-on-conn whose fiber has already terminated. Throw a protocol error to initiate a shutdown to prevent `RequestOnConn.setReceivedPayload` from attempting to resume the terminated fiber.

Fixes #309.

### DIFF
--- a/src/swarm/neo/client/RequestOnConnSet.d
+++ b/src/swarm/neo/client/RequestOnConnSet.d
@@ -135,6 +135,7 @@ public struct RequestOnConnSet
     {
         this.list ~= request_on_conn;
         this.num_active++;
+        request_on_conn.active = true;
 
         return request_on_conn;
     }
@@ -167,6 +168,7 @@ public struct RequestOnConnSet
             "request-on-connection already exists for the specified node");
 
         this.num_active++;
+        request_on_conn.active = true;
 
         return request_on_conn;
     }
@@ -250,20 +252,29 @@ public struct RequestOnConnSet
 
     /***************************************************************************
 
-        Decrements the `num_active` counter and returns whether it is now 0.
+        Registers `roc` as inactive, decrements the `num_active` counter and
+        returns whether the counter is now 0.
+
+        Params:
+            roc = the `RequestOnConn` instance that has finished
 
         Returns:
             true if `num_active` is 0
 
     ***************************************************************************/
 
-    public bool finished ( )
+    public bool finished ( RequestOnConn roc )
     in
     {
         assert(this.num_active);
     }
     body
     {
+        // TODO: Consider `this.all_nodes.remove(roc)` instead of using the
+        // `active` flag. Not doing this now for a patch release to avoid
+        // potentially introducing a bug if there is an unobvious reason why it
+        // isn't done.
+        roc.active = false;
         return --this.num_active == 0;
     }
 
--- a/src/swarm/neo/client/RequestSet.d
+++ b/src/swarm/neo/client/RequestSet.d
@@ -439,7 +439,7 @@ public final class RequestSet: IRequestSet
         }
         body
         {
-            if ( this.getRequestOnConnForNode(connection.remote_address) !is null )
+            if ( this.request_on_conns.get(connection.remote_address) !is null )
                 return;
 
             auto request_on_conn = this.request_on_conns.add(
@@ -452,21 +452,37 @@ public final class RequestSet: IRequestSet
         /***********************************************************************
 
             Obtains the handler of this request that is currently communicating
-            with a node, if any.
+            with a node, if any. Expects the handler to be active and throws `e`
+            if not.
 
             Params:
                 node_address = the address of the node to look up the handler
                                for
+                e = the exception to throw if a handler was found that is not
+                    active any more but has finished
 
             Returns:
                 the handler of this request that is currently communicating with
                 the node or null if none is doing that right now.
 
+            Throws:
+                `ProtocolError` if a handler was found that is not active any
+                more but has finished.
+
         ***********************************************************************/
 
-        public RequestOnConn getRequestOnConnForNode ( AddrPort node_address )
+        public RequestOnConn getRequestOnConnForNode ( AddrPort node_address,
+            ProtocolError e )
         {
-            return this.request_on_conns.get(node_address);
+            if (auto roc = this.request_on_conns.get(node_address))
+            {
+                if (roc.active)
+                    return roc;
+                else
+                    throw e.set("Unexpected incoming message");
+            }
+            else
+                return null;
         }
 
         /***********************************************************************
@@ -617,7 +633,7 @@ public final class RequestSet: IRequestSet
 
         public void handlerFinished ( RequestOnConn request_on_conn )
         {
-            if (this.request_on_conns.finished())
+            if (this.request_on_conns.finished(request_on_conn))
             {
                 scope working_data_iter = new RequestWorkingData;
                 this.finished_notifier(this.context, working_data_iter);
--- a/src/swarm/neo/client/Connection.d
+++ b/src/swarm/neo/client/Connection.d
@@ -602,7 +602,8 @@ public final class Connection: ConnectionBase
     {
         if (auto request = this.request_set.getRequest(id))
         {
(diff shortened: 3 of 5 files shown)


src/swarm/neo/client/RequestOnConnSet.d
```diff
@@ -252,25 +252,29 @@ public struct RequestOnConnSet
 
     /***************************************************************************
 
-        Decrements the `num_active` counter and returns whether it is now 0.
+        Decrements the `num_active` counter, registers `roc` as inactive and
+        returns whether it is now 0.
```
gavin-norman-sociomantic: Your addition to this sentence has changed the meaning of "it" at this point.

src/swarm/neo/client/RequestOnConnSet.d
```diff
@@ -252,25 +252,29 @@ public struct RequestOnConnSet
...
+
+        Params:
+            roc = the `RequestOnConn` instance to register as inactive
```
gavin-norman-sociomantic: Maybe "the `RequestOnConn` instance that has finished" is a bit clearer?

src/swarm/neo/client/RequestOnConnSet.d
```diff
@@ -252,25 +252,29 @@ public struct RequestOnConnSet
...
         assert(this.num_active);
     }
     body
     {
-        // TODO: Consider `this.all_nodes.remove(request_on_conn)` instead of
-        // using the `active` flag. Not doing this now for a patch release to
-        // avoid potentially introducing a bug if there is an unobvious reason
-        // why it isn't done.
-        request_on_conn.active = false;
+        // TODO: Consider `this.all_nodes.remove(roc)` instead of using the
```
gavin-norman-sociomantic: Actually, this would be better as an issue, rather than a code comment. We'll probably never think of it again, if it's left as a comment deep in the code like this.

src/swarm/neo/client/RequestOnConnSet.d
```diff
@@ -264,6 +266,11 @@ public struct RequestOnConnSet
     }
     body
     {
+        // TODO: Consider `this.all_nodes.remove(request_on_conn)` instead of
+        // using the `active` flag. Not doing this now for a patch release to
+        // avoid potentially introducing a bug if there is an unobvious reason
+        // why it isn't done.
+        request_on_conn.active = false;
```
gavin-norman-sociomantic: What does `request_on_conn` refer to? Looks like it's intended to be an argument, but the function has no arguments.

src/swarm/neo/client/RequestSet.d
```diff
@@ -439,7 +439,8 @@ public final class RequestSet: IRequestSet
         }
         body
         {
-            if ( this.getRequestOnConnForNode(connection.remote_address) !is null )
+            if ( this.getRequestOnConnForNode(connection.remote_address,
```
gavin-norman-sociomantic: Oh this is a weird case! This should be impossible, right? (If so, this is more like a `verify` than a protocol error.)

gavin-norman-sociomantic (commented):
It'd be nice for the commit message to include a high-level summary of what the old and behaviour is, in this scenario.

src/swarm/neo/client/RequestSet.d
david-eckardt-sociomantic: Passing the exception here is merely for the sake of the arguments of `getRequestOnConnForNode`. I can add a `null` default for the exception parameter.

david-eckardt-sociomantic:
Updated, addressing @gavin-norman-sociomantic's comments. Added issue #311.

gavin-norman-sociomantic (approved):

PR #314, opened 2018-03-19
Add missing import for ProtocolError

### LINKED ISSUE #313: swarm v5.0.8 doesn't compile with dmd2.0.71 and dmdqproto14.1.1
Does not compile with dmqproto v14.1.1 and dmd v2.0.71

./submodules/swarm/src/swarm/neo/client/RequestSet.d(430): Error: undefined identifier 'ProtocolError'

nemanja-boric-sociomantic:
Add the import to the ProtocolError, since it's used in the RequestSet class.

Fixes #313

### DIFF
--- a/src/swarm/neo/client/RequestSet.d
+++ b/src/swarm/neo/client/RequestSet.d
@@ -40,6 +40,7 @@ public final class RequestSet: IRequestSet
     import swarm.neo.client.RequestOnConnSet;
     import swarm.neo.client.RequestHandlers;
     import swarm.neo.connection.YieldedRequestOnConns;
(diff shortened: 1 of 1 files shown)


don-clugston-sociomantic:
LGTM

gavin-norman-sociomantic:
It's not clear why this code didn't fail in swarm / *proto before. Investigating...

gavin-norman-sociomantic:
The reason is that swarm v4.6.x is not being compiled with dmd 2.071. So changes made in v4.6.x were not checked for 2.071 compatibility until they were merged into v5.0.x, which is tested with 2.071.

PR #315, opened 2018-03-29
Enable TCP_NODELAY for the client before adding nodes

nemanja-boric-sociomantic:
In case the NeoConfig enabled constructor is used, the TCP_NODELAY
can't be set downstream, because it needs to be set before the nodes
are added, and the swarm user is not able to do that if the nodes are
added in the constructor. This simply turns the TCP_NODELAY on when
constructing the neo client.

Adding it here to fix the issue in swarm v4.6.x/v5.0.x - this behaviour is on by default in the next,
not yet released minor.

### DIFF
--- a/src/swarm/neo/client/mixins/ClientCore.d
+++ b/src/swarm/neo/client/mixins/ClientCore.d
@@ -239,6 +239,7 @@ template ClientCore ( )
             settings.conn_notifier, settings.auto_connect);
         this.request_resources = settings.request_resources;
 
+        this.enableSocketNoDelay();
         this.addNodes(config.nodes_file());
     }
 


nemanja-boric-sociomantic:
cc @jens-mueller-sociomantic

src/swarm/neo/client/mixins/ClientCore.d
```diff
@@ -161,6 +161,7 @@ template ClientCore ( )
         this.connections = new ConnectionSet(cred, this.outer.epoll,
             settings.conn_notifier, settings.auto_connect);
         this.request_resources = settings.request_resources;
+        this.enableSocketNoDelay();
```
gavin-norman-sociomantic: This case isn't mentioned in the commit message. What's the reasoning for doing it here as well?

nemanja-boric-sociomantic: It has to be done before the `Connections` inside `ConnectionSet` are created - since it's passed via constructor of every `Connection` object. Was this not clear enough - do you prefer more detailed explanation?

> In case the NeoConfig enabled constructor is used, the TCP_NODELAY
can't be set downstream, because it needs to be set before the nodes
are added [...]

gavin-norman-sociomantic: It's just that the commit message says "In case the NeoConfig enabled constructor is used". That implies just the 2nd ctor, or?

nemanja-boric-sociomantic: Ah, you're right. I'll reword.

nemanja-boric-sociomantic: Reworded.

gavin-norman-sociomantic (approved):

PR #316, opened 2018-04-03
Replace usage of getMsg with Exception.message

mathias-lang-sociomantic:
The former always resolves to the later nowadays.

### DIFF
--- a/src/swarm/README_client_neo.rst
+++ b/src/swarm/README_client_neo.rst
@@ -174,7 +174,7 @@ delegate looks something like this imaginary example:
 
       case error: // example I/O error notification
         Stderr.formatln("Request on channel {} failed due to error {} "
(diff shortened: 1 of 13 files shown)


gavin-norman-sociomantic (approved):

gavin-norman-sociomantic:
Can confirm that swarm v4.x.x is using a runtime where `Exception.message` exists. (The [dockerfile](https://github.com/sociomantic-tsunami/swarm/blob/v4.x.x/beaver.Dockerfile#L1) uses [cachalot image v2](https://github.com/sociomantic-tsunami/cachalot/blob/v2.0.0/dlang.Dockerfile), which uses [tangort v1.7.x](https://github.com/sociomantic-tsunami/tangort/blob/v1.7.0/src/rt/object_.d#L1365).)

PR #320, opened 2018-04-30
Verify the RequestSetup module

nemanja-boric-sociomantic:
This module was slipped while converting asserts into verify calls.


gavin-norman-sociomantic (approved):

gavin-norman-sociomantic:
Oh! I wonder how we missed this!

nemanja-boric-sociomantic:
The entire file was not covered, so it's probably just slip. I've double checked, and it seems that all other files are covered.

PR #323, opened 2018-05-15
Make neo RequestContext getters const

nemanja-boric-sociomantic:
Since we're passing the const object of RequestContext to the user,
getters needs to be marked as const to be usable.

### DIFF
--- a/src/swarm/neo/client/request_options/RequestContext.d
+++ b/src/swarm/neo/client/request_options/RequestContext.d
@@ -150,7 +150,7 @@ public struct RequestContext
 
     ***************************************************************************/
 
-    public ulong integer ( )
+    public ulong integer ( ) /* d1to2fix_inject: const */
     {
         return this.context().integer;
     }
@@ -165,7 +165,7 @@ public struct RequestContext
 
     ***************************************************************************/
 
-    public Object object ( )
+    public Object object ( ) /* d1to2fix_inject: const */
     {
         return this.context().object;
     }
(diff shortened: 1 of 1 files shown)


src/swarm/neo/client/request_options/RequestContext.d
```diff
@@ -165,7 +165,7 @@ public struct RequestContext
 
     ***************************************************************************/
 
-    public Object object ( )
+    public Object object ( ) /* d1to2fix_inject: const */
```
nemanja-boric-sociomantic, line 168: The core of my problem: strictly speaking this is a breaking change...

gavin-norman-sociomantic: How does this break the library API?

nemanja-boric-sociomantic: Ha, this is interesting - it doesn't! The reason why I thought it would break API is I assumed that the `Object` returned would be `const Object` (since it's the part of the `RequestContext` object, and this method is marked `const`, so the `RequestContext` should not change). However, there's a carefuly placed `cast` here, so this all works (although not sure how correct it is).

In other words:

```D
class X
{
    import ocean.core.ContextUnion;

    private ubyte[ContextUnion.sizeof] context_;
    private Object o;

    Object get() const // compiles
    {
        return (cast(ContextUnion*)this.context_.ptr).object();
    }

    Object getMe() const // doesn't compile - can't cast this.o of type `const(Object)` to `Object`
    {
        return this.o;
    }
}
```

gavin-norman-sociomantic: Ha, interesting. All good then.

nemanja-boric-sociomantic:
Oh, this should actually hit v4.7.x.

nemanja-boric-sociomantic:
Rebased.

gavin-norman-sociomantic (approved):

PR #325, opened 2018-05-23
Add closure allocation check to CI

mihails-strasuns-sociomantic:


integrationtest/neo/client/request/internal/DoublePut.d
```diff
@@ -150,69 +150,69 @@ public struct DoublePut
             start_request_on_new_conn();
 
         // Send to the selected node.
-        use_node(node,
```
gavin-norman-sociomantic, line 153: I don't think it's worth fixing closure allocations in tests.

src/swarm/neo/client/RetryTimer.d
```diff
@@ -98,8 +85,14 @@ body
         // Constructor; sets the timer to the next delay that is greater
         // than the duration of the first retry, `t_end - t_start`, and
         // registers this instance with epoll.
-        this ( )
+        this ( timespec t_start, timespec t_end )
         {
+            // This exception is caught inside the scope of this function.
+            this.e = new TimerFD.TimerException;
```
gavin-norman-sociomantic: These objects are now heap allocated. (They were previously stack allocated.)

gavin-norman-sociomantic (dismissed):

integrationtest/neo/client/request/internal/DoublePut.d
mihails-strasuns-sociomantic: We discussed it - I can't distinguish between tests and non tests, this is the price of making the check non-ignorable. It is somewhat annoying but not that much.

src/swarm/neo/client/RetryTimer.d
mihails-strasuns-sociomantic: Oh, this is going to be interesting. Means only way to avoid closure is to use something like Phobos `emplace` (https://dlang.org/library/std/conv/emplace.html#2) :/

nemanja-boric-sociomantic: We do have `emplace` inside swarm (we allocate request handlers on stack using it): https://github.com/sociomantic-tsunami/swarm/blob/6fa8a821aa6cb3bfbe00c61b9652aa0df43d2a3d/src/swarm/neo/node/ConnectionHandler.d#L691

nemanja-boric-sociomantic: (it's a bit primitive atm, though, as it doesn't accept constructor arguments)

mihails-strasuns-sociomantic: I think we should add few utils like that to ocean considering how widespread this may end up being

gavin-norman-sociomantic: Without looking at the code: afair, there's no real reason for this code to be all nests within nests, like it's currently written. Could it be refactored to avoid this problem completely?

mihails-strasuns-sociomantic: Yeah I am going to do it tomorrow

mihails-strasuns-sociomantic: @gavin-norman-sociomantic I looked at it again and the whole stack allocation thing looks rather pointless to me here. For example, `TimerFD` constructor will heap allocate exception instance internally - it doesn't matter that you create it as `scope t = new TimerFD`, there will be GC activity anyway.

Same applies for few other fields - and considering usually GC is "all or nothing" for our code, it doesn't look like this is optimized as it claims to be.

gavin-norman-sociomantic: Cool. Ideally without using nasty tricks like `emplace` ;)

mihails-strasuns-sociomantic: See updated version - I have simply moved class to global scope and made it refer scope-allocated stuff as reference captured by constructor.

nemanja-boric-sociomantic:
> @gavin-norman-sociomantic I looked at it again and the whole stack allocation thing looks rather pointless to me here. For example, TimerFD constructor will heap allocate exception instance internally - it doesn't matter that you create it as scope t = new TimerFD, there will be GC activity anyway.

That's a pretty good point. For some reason I believed that the `scope` was transitive.

src/swarm/neo/client/RetryTimer.d
```diff
@@ -62,139 +62,159 @@ body
...
-                    // Set the timer to expire every 2.5s.
-                    timer.set(delay[this.n], delay[this.n]);
-            }
+*******************************************************************************/
 
-            epoll.register(this);
-        }
+private final class TimerEvent: ISelectClient
+{
+    // externally created timer reference
```
gavin-norman-sociomantic: Could you standardise comment formatting throughout this class? (`///` / `/**`)

mihails-strasuns-sociomantic:
> That's a pretty good point. 

Nevermind that (I deleted the comment) - old code was creating exception field separately on scope for exactly that reason. I reproduced it in last version.

ci/closures.sh
```diff
@@ -0,0 +1,11 @@
...
+set -xe
+
+# Enables printing of all potential GC allocation sources to stdout
+export DFLAGS=-vgc
+
+# Prepare sources
+beaver dlang make d2conv
+
+# Ensure that there are no lines about closure allocations in the output
+! beaver dlang make fasttest 2>&1 | grep "closure"
```
leandro-lucarella-sociomantic: Only fast because this is detected at compile-time, right? So we only need to parse the actual library modules and not any unit-tests modules?

ci/closures.sh
```diff
@@ -8,4 +8,5 @@ export DFLAGS=-vgc
 beaver dlang make d2conv
 
 # Ensure that there are no lines about closure allocations in the output
-! beaver dlang make fasttest 2>&1 | grep "closure"
+# Filters away errors from submodules to focus on swarm own code only
+! beaver dlang make fasttest 2>&1 | grep -v "submodule" | grep "closure"
```
leandro-lucarella-sociomantic: Mmm, not crazy that the actual `.gitmodules` config is not involved in this. What about grepping for `$(git rev-parse --show-toplevel)/src` instead?

leandro-lucarella-sociomantic: Also is one `grep` less:
```sh
! beaver dlang make fasttest 2>&1 | grep "$(git rev-parse --show-toplevel)/src/.*\<closure\>"
```

relnotes/closures.feature.md
```diff
@@ -0,0 +1,5 @@
+### Fixup of swarm internals to prevent closure allocation
+
+Some of internal swamr code was refactored so that it doesn't allocate closures
```
leandro-lucarella-sociomantic: `swarm` typo

ci/closures.sh
mihails-strasuns-sociomantic: Correct - and we get less false positives this way (we don't really care about closures in tests)

mihails-strasuns-sociomantic:
Updated - not WIP anymore:

- Documentation tweak as requested by @gavin-norman-sociomantic 
- Regexp improvement as suggested by @leandro-lucarella-sociomantic

ci/closures.sh
mihails-strasuns-sociomantic: Hm, no, this is wrong - compiler output will have relative paths, not absolute ones. Need to tweak this..

mihails-strasuns-sociomantic: Updated

mihails-strasuns-sociomantic:
Ping @leandro-lucarella-sociomantic @gavin-norman-sociomantic

relnotes/closures.feature.md
```diff
@@ -0,0 +1,5 @@
+### Fixup of swarm internals to prevent closure allocation
```
gavin-norman-sociomantic: A feature release note in a patch release?

mihails-strasuns-sociomantic: Yes, we do it occasionally when fixing something implies adding new feature behind the scene. Though here maybe it is more appropriate to just mark it as `bug.md`

gavin-norman-sociomantic: Right, that's what I thought. `### Fixup` is the hint ;)

mihails-strasuns-sociomantic: Updated

gavin-norman-sociomantic: Mm, still looks like ` relnotes/closures.feature.md` to me.

nemanja-boric-sociomantic: I don't see this as a feature, but as a bugfix for two reasons:

- it doesn't impose anything to the user API-wise
- it makes swarm to work as expected.

I don't think this actually needs any release note.

mihails-strasuns-sociomantic:
Ping

gavin-norman-sociomantic:
Ah I hadn't noticed that you'd fixed the release notes.

gavin-norman-sociomantic (approved):

PR #337, opened 2018-06-25
Log IOError node exceptions with info level.

nemanja-boric-sociomantic:
If the client ever dies during the handling of the request (which
is amplified by the long-living requests), `IOError` will be thrown
in the node and it will be logged as an error. This is not an error
from the node's PoV and it's logged now with an info level.

### DIFF
--- a/src/swarm/neo/node/ConnectionHandler.d
+++ b/src/swarm/neo/node/ConnectionHandler.d
@@ -55,6 +55,7 @@ class ConnectionHandler : IConnectionHandler
     import swarm.neo.request.Command;
 
     import ocean.io.select.EpollSelectDispatcher;
+    import ocean.io.select.protocol.generic.ErrnoIOException;
     import ocean.time.StopWatch;
 
     import ClassicSwarm =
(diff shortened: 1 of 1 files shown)


gavin-norman-sociomantic (approved):
Thanks! This will help in turtle, right?

Btw "amplituted" -> "amplified"?

nemanja-boric-sociomantic:
Yes, "amplified" ;-). It helps mainly on the live nodes, since we don't want to generate alerts for logged errors which are not errors.

nemanja-boric-sociomantic:
Updated commit message.

nemanja-boric-sociomantic:
(but it also helps in turtle, yes)

gavin-norman-sociomantic:
Good deed of the day: help a turtle :heavy_check_mark:

nemanja-boric-sociomantic:
I'll merge this tomorrow to prevent broken Travis builds after merging (due to sourceforge's outage).

gavin-norman-sociomantic:
Are you saying I should merge it before then? Or that we must wait until then?

nemanja-boric-sociomantic:
We should wait until then. If we merge this, the travis will trigger v4.6.x branch rebuild which will fail due to sourceforge outage, so we'll have a red branch.

nemanja-boric-sociomantic:
I think we can merge this now.

PR #341, opened 2018-06-29
Remove internal deprecations.

nemanja-boric-sociomantic:
This PR removes references to deprecated methods in swarm from unittests, so swarm doesn't hit it's own deprecations.

### DIFF
--- a/src/swarm/neo/client/mixins/ClientCore.d
+++ b/src/swarm/neo/client/mixins/ClientCore.d
@@ -239,7 +239,6 @@ template ClientCore ( )
             settings.conn_notifier, settings.auto_connect);
         this.request_resources = settings.request_resources;
 
-        this.enableSocketNoDelay();
         this.addNodes(config.nodes_file());
     }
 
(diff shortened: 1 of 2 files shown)


gavin-norman-sociomantic:
Yeah, I was going to say, why undeprecate it when we can just not call it internally. I see you updated ;)

nemanja-boric-sociomantic:
Well, the `protos` still need to call it. However, I forgot that we can do `static if` on the swarm version inside protos.

gavin-norman-sociomantic:
The protos need to call it because in older swarm versions it used to do something?

nemanja-boric-sociomantic:
Yes, in older swarm versions, TCP_NODELAY was not turned on by default and we thought that's too much for a patch release.

src/swarm/neo/client/mixins/ClientCore.d
```diff
@@ -239,7 +239,6 @@ template ClientCore ( )
             settings.conn_notifier, settings.auto_connect);
```
gavin-norman-sociomantic, line 239: >This method is de-facto deprecated and does nothing.

Mini comment: this is wrong. The method is _actually_ deprecated.

nemanja-boric-sociomantic: Thanks, updated.

gavin-norman-sociomantic (approved):

PR #343, opened 2018-07-16
Sanitize the input array length

nemanja-boric-sociomantic:
FiberSelectReader now will throw an exception if the array
length is larger than FiberSelectReader.max_array_size (defaults
to 10MB). This will prevent allocation of the huge amounts of memory
on the invalid data.

### DIFF
--- a/relnotes/readarray.bug.md
+++ b/relnotes/readarray.bug.md
@@ -1,0 +1,6 @@
+### Validate the length of the array before reading it
+
+* `swarm.protocol.FiberSelectReader`
+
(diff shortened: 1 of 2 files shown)


gavin-norman-sociomantic (changes requested):
Shouldn't update beaver and makd, or?

nemanja-boric-sociomantic:
Whops, that's a mistake.

nemanja-boric-sociomantic:
Updated.

gavin-norman-sociomantic (approved):

PR #344, opened 2018-07-16
Set max_array_size from the second constructor in FiberSelectReader

nemanja-boric-sociomantic:
This field was set only from the first constructor, failing to
set it from the second, making it zero.


nemanja-boric-sociomantic:
Sorry :angel:.

nemanja-boric-sociomantic:
After merging, I'll try it in DHT node before releasing (it works locally).

nemanja-boric-sociomantic:
Green.

gavin-norman-sociomantic (approved):

PR #347, opened 2018-07-26
[WIP] Adapt turtle env to allow clusters of nodes, not just a single node

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Super rough sketch, to be played with in a proto repo (where most of the work will be).

Issue #348, opened 2018-07-27
Replace internal usage of NodeItem with AddrPort


gavin-norman-sociomantic:
A mix of the two structs is used internally, which is really inconvenient.

gavin-norman-sociomantic:
This is a big task. https://github.com/sociomantic-tsunami/swarm/pull/353 scratches the surface.

I think we should delay this and do it as part of https://github.com/sociomantic-tsunami/swarm/issues/57. It's not worth trying to unravel things right now.

gavin-norman-sociomantic:
>I think we should delay this and do it as part of #57. It's not worth trying to unravel things right now.

Me in December agrees with me in August. Closing.

PR #350, opened 2018-07-27
Change use of NodeItem to AddrPort in node base classes

gavin-norman-sociomantic:


gavin-norman-sociomantic:
Note to reviewer: could you confirm the changes around `InetAddress` and `AddrPort`? It's quite complicated.

gavin-norman-sociomantic:
(I was originally intending to replace all internal usage of `NodeItem` with `AddrPort` -- see https://github.com/sociomantic-tsunami/swarm/issues/348 -- but this is a massive undertaking, as it's used extensively in the client. This PR just updates a small section -- the node bases.)

src/swarm/node/model/INodeInfo.d
```diff
@@ -38,9 +39,20 @@ public interface INodeInfo
 
      **************************************************************************/
 
+    deprecated("Use addr_port instead")
     public NodeItem node_item ( );
```
nemanja-boric-sociomantic, line 43: Deprecation node?

src/swarm/node/model/INodeInfo.d
```diff
@@ -38,9 +39,20 @@ public interface INodeInfo
...
 
 
+    /**************************************************************************
+
+        Returns:
+            NodeItem struct containing the address & port of the node
+
+     **************************************************************************/
+
+    public AddrPort addr_port ( );
```
nemanja-boric-sociomantic, line 53: Breaking change?

gavin-norman-sociomantic:
Blocked pending testing with proto and node repos.

src/swarm/node/model/NeoNode.d
```diff
@@ -845,13 +891,18 @@ public class NodeBase ( ConnHandler : ISwarmConnectionHandler ) : INodeBase
         }
 
         // Super ctor.
+        InetAddress!(false) addr, neo_addr;
+        addr = cast(sockaddr_in)node;
```
nemanja-boric-sociomantic, line 896: Consider adding `opCast` from `AddrPort` directly to `InetAddr`.

src/swarm/node/model/NeoNode.d
```diff
@@ -845,13 +891,18 @@ public class NodeBase ( ConnHandler : ISwarmConnectionHandler ) : INodeBase
...
+        neo_addr = addr;
+        neo_addr.port = neo_port;
+
         super(node, conn_setup_params,
             this.listener = new Listener(
-                addr(node.Address, node.Port), this.socket, conn_setup_params,
+                cast(sockaddr*)&addr.addr, this.socket, conn_setup_params,
```
nemanja-boric-sociomantic, line 902: This is fine, considering that `addr.addr` is `sockaddr_in`. However, I think by casting pointer we're loosing the type safety here. I would just add the static assert before `static assert(typeof(addr.addr) is sockaddr_in)`.

src/swarm/node/model/INodeInfo.d
gavin-norman-sociomantic: Could you elaborate?

gavin-norman-sociomantic:
On second thoughts, I'd prefer to not address this now. We plan to completely rewrite all of the core node stuff, anyway, at some point, so it seems better to leave it alone until then.

Replaced with #354, #355.

PR #365, opened 2018-10-09
Add suspend unregister callback

matthias-wende-sociomantic:
This may be used in any proto to give applications an option to
remove instances of type ISuspendable when a suspendable request
finishes.

### DIFF
--- a/src/swarm/client/RequestSetup.d
+++ b/src/swarm/client/RequestSetup.d
@@ -15,6 +15,8 @@
 
 module swarm.client.RequestSetup;
 
+import ocean.core.Traits;
+
 import swarm.client.request.params.IChannelRequestParams;
 
 /*******************************************************************************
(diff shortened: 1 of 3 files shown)


gavin-norman-sociomantic:
High-level question: is this an important bug fix? Our general principle with all legacy swarm code is to not change anything unless it's very important.

matthias-wende-sociomantic:
That is part of the "proper" fix to the problem described in https://github.com/sociomantic-tsunami/ocean/pull/640 (i.e. applications should remove a suspendable from the list of suspendables when a request finishes rather then keeping it in the list in a undefined state). I would consider this as a bug fix feature :).

matthias-wende-sociomantic:
Alright, as discussed IRL we not gonna merge this since the switch to neo is near ;).

src/swarm/client/RequestSetup.d
```diff
@@ -31,7 +31,7 @@ public template RequestParamsSetup ( )
         Imports needed by template.
```
gavin-norman-sociomantic, line 33: I wouldn't mention dhtproto in the commit message specifically. This change could, in principle, be used in any proto repo.

gavin-norman-sociomantic (changes requested):

gavin-norman-sociomantic:
Oh this should have a release note too (new feature).

nemanja-boric-sociomantic (approved):

gavin-norman-sociomantic (approved):

PR #366, opened 2018-10-09
Verify that neo client config instance is non-null before using it

joseph-wakeling-sociomantic:
This patch adds a small but useful sanity check to the constructor for neo client support.  This should ensure that failure to instantiate the neo client config class properly results in an informative exception and error message rather than a segfault.


gavin-norman-sociomantic (approved):

gavin-norman-sociomantic:
Merci!

PR #367, opened 2018-10-24
Implement ISuspendable.suspended in Controller's suspendable

nemanja-boric-sociomantic:
Previously, the controller's suspendable didn't implement the
ISuspendable interface because there's no reliable way to do so.
This is now amended in a best-effort strategy: the ISuspendable will
report if the stream was suspended/resumed through the interface itself.

### DIFF
--- a/src/swarm/neo/client/mixins/Controllers.d
+++ b/src/swarm/neo/client/mixins/Controllers.d
@@ -191,6 +191,20 @@ template Controllers ( )
 
         private Pending pending;
 
+
+        /***********************************************************************
+
+            Internal state of this ISuspendable. Note that the this indicates the
+            last action done through this suspendable, and it may not reflect
+            the actual reality (the streams may need time to suspend). This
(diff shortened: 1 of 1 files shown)


src/swarm/neo/client/mixins/Controllers.d
```diff
@@ -191,6 +191,20 @@ template Controllers ( )
 
         private Pending pending;
 
+
+        /***********************************************************************
+
+            Internal state of this ISuspendable. Note that the this indicates the
+            last action done through this suspendable, and it may not reflect
+            the actual reality (the streams may need time to suspend). This
+            implies that one can't control stream directly through the controller
```
gavin-norman-sociomantic, line 200: Yeah, the whole `Suspendable` class was always regarded as a temporary hack. (The doc above says: "In the future, we may adapt the suspendable/throttling code in ocean to handle this different case.")

I'd like to reconsider the whole problem before patching the hack.

PR #371, opened 2018-11-06
Don't wait for the connection to be reestablished for single node requests

nemanja-boric-sociomantic:
In case when the connection was never established to the node, all
single-node requests will fail, because the connection will not be found
in the connection set. However, in case when the connection breaks
during the lifetime of the client, the client will keep scheduling
requests for the connection until the connection is reeastablished.

This change checks the connection status before starting the single-node
request. In case the connection is not established, the client report the
same error as if the connection was never established.

### DIFF
--- a/src/swarm/neo/client/RequestOnConn.d
+++ b/src/swarm/neo/client/RequestOnConn.d
@@ -782,6 +782,11 @@ public class RequestOnConn: RequestOnConnBase, IRequestOnConn
         if (!this.connection)
             return false;
 
+        auto client_connection = cast(Connection)this.connection;
+        assert(client_connection);
+        if (client_connection.status != Connection.Status.Connected)
+            return false;
+
         scope (exit)
         {
             // the request may decide to use another connection, so we need to


joseph-wakeling-sociomantic:
I've tried this with the app where we spotted the hanging put requests, and it seems to work: where previously we saw DHT.neo `Put` requests just never completing (with the notifier never being callled), now we see those `Put` requests explicitly failing as expected.

src/swarm/neo/client/RequestOnConn.d
```diff
@@ -782,6 +782,11 @@ public class RequestOnConn: RequestOnConnBase, IRequestOnConn
         if (!this.connection)
             return false;
 
+        auto client_connection = cast(Connection)this.connection;
```
gavin-norman-sociomantic, line 785: Looking at the way the [DHT Put request](https://github.com/sociomantic-tsunami/dhtproto/blob/neo-v0.x.x/src/dhtproto/client/request/internal/Put.d#L133-L145) (for example) is implemented, this change would cause the wrong notification to be sent to the user, as far as I can see.

`handler` calls [`NodeHashRanges.putToNode`](https://github.com/sociomantic-tsunami/dhtproto/blob/neo-v0.x.x/src/dhtproto/client/internal/NodeHashRanges.d#L192), which calls `useNode` (the one changed in this PR). If the connection is not up, `useNode` now returns without calling the provided delegate. This causes `handler` to set the result status to `NoNode` (as `put_called` was not set), which causes the `no_node` notification.

We'd actually expect `node_disconnected` to happen, in this situation.

gavin-norman-sociomantic: Not saying that we shouldn't make this fix, but we might want to investigate the knock-on effects of the change a bit more deeply. It looks like it could lead to pretty subtle and unexpected changes in behaviour.

gavin-norman-sociomantic (approved):
We investigated the effects that this change will have on the behaviour of all extant single-node requests:
* It may cause slightly odd behaviour (the wrong type of error notifications being called).
* It should not cause any seriously wrong behaviour.

So we decided to go ahead with the change. We'll do a proper fix for this in the next major (it's a breaking change): https://github.com/sociomantic-tsunami/swarm/issues/372

src/swarm/neo/client/RequestOnConn.d
```diff
@@ -782,6 +782,11 @@ public class RequestOnConn: RequestOnConnBase, IRequestOnConn
...
+        assert(client_connection);
```
daniel-zullo, line 786: I wonder if this assertion is redundant. Wouldn't the method return when checking `if (!this.connection)`?

nemanja-boric-sociomantic: I'm here checking if the cast succeeded, though.

daniel-zullo: right, I thought it wasn't needed

PR #380, opened 2018-12-20
Add protocol documentation section outlines

gavin-norman-sociomantic:


gavin-norman-sociomantic:
I'd suggest we just merge this as is. Having an outline is better than nothing.

matthias-wende-sociomantic:
Agree!

matthias-wende-sociomantic (approved):

PR #381, opened 2018-12-27
Validate node ip address at construction time

gavin-norman-sociomantic:


src/swarm/Const.d
```diff
@@ -115,6 +122,29 @@ public struct NodeItem
...
+
+        Returns:
+            true if the Address field is a valid IPv4 address
+
+    ***************************************************************************/
+
+    public bool isAddressValidIPv4 ( )
+    {
+        InetAddress!() check_ip;
+        return check_ip.inet_pton(this.Address) == 1;
```
matthias-wende-sociomantic: Why not check IPv6 at the same time? IIUC it can be checked using the `inet_pton` method from the same `InetAddress` struct when constructing it with a `true` template parameter.

gavin-norman-sociomantic: Yeah I wondered about doing that. The reason I didn't is that nothing in swarm officially supports IPv6, afair. What do you think?

matthias-wende-sociomantic: As it is almost trivial and it doesn't harm to have it, I don't see any reason not to do it . After all it's almost 2020  :smile:.

gavin-norman-sociomantic: So what would the behaviour be on parsing a valid IPv6 address? Nothing else in swarm officially supports IPv6, so it would have to reject it I guess. Which is the equivalent to what we have now in this PR ;)

gavin-norman-sociomantic: To be honest, I have no idea how much work would be needed in swarm/ocean to support IPv6 (or to check whether it's supported), and don't really want to open that can of worms right now.

gavin-norman-sociomantic:
Updated.

matthias-wende-sociomantic (approved):

PR #385, opened 2019-01-03
Rework channel creation code in IStorageChannels

gavin-norman-sociomantic:

PR #387, opened 2019-01-15
Allow NodeBase derivatives to access the unix socket command handlers set

gavin-norman-sociomantic:
Derived classes should be able to add their own command handlers. Previously, it was only possible for the base class to add handlers.

### DIFF
--- a/src/swarm/node/model/NeoNode.d
+++ b/src/swarm/node/model/NeoNode.d
@@ -728,6 +728,14 @@ public class NodeBase ( ConnHandler : ISwarmConnectionHandler ) : INodeBase
 
     /***************************************************************************
 
+        Unix socket command handlers map.
+
+    ***************************************************************************/
+
+    protected BasicCommandHandler.Handler[istring] unix_socket_handlers;
+
+    /***************************************************************************
+
         The server socket.
 
     ***************************************************************************/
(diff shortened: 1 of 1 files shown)


gavin-norman-sociomantic:
Cherry picked from https://github.com/sociomantic-tsunami/swarm/pull/385.

nemanja-boric-sociomantic (approved):

src/swarm/node/model/NeoNode.d
```diff
@@ -813,16 +821,15 @@ public class NodeBase ( ConnHandler : ISwarmConnectionHandler ) : INodeBase
...
-                unix_socket_handlers =
+                this.unix_socket_handlers =
                     ["update-credentials"[]: &this.handleUpdateCredentials,
                      "list-credentials": &this.handleListCredentials];
             }
 
-            unix_listener = new UnixListener(
-                options.unix_socket_path, options.epoll, unix_socket_handlers);
+            unix_listener = new UnixListener(options.unix_socket_path,
+                options.epoll, this.unix_socket_handlers);
```
nemanja-boric-sociomantic, line 832: I think `UnixListener` doesn't make a copy of the request handlers, but keeps a reference to these, so it's safe to do this (because in derived class, you would first create the handler then adjust the `BasicCommandHandler` array).

src/swarm/node/model/NeoNode.d
```diff
@@ -726,6 +726,14 @@ public class NodeBase ( ConnHandler : ISwarmConnectionHandler ) : INodeBase
     private NeoListener neo_listener;
 
 
+    /***************************************************************************
+
+        Unix socket command handlers map.
+
+    ***************************************************************************/
+
+    protected BasicCommandHandler.Handler[istring] unix_socket_handlers;
```
nemanja-boric-sociomantic, line 735: This could be an method, but that might be too much.

src/swarm/node/model/NeoNode.d
gavin-norman-sociomantic: You're confirming that my change is correct, right?

src/swarm/node/model/NeoNode.d
gavin-norman-sociomantic: To prevent derived classes from modifying the reference? Yeah... I guess that's not the standard approach in swarm.

nemanja-boric-sociomantic: Yeah I'm fine with this as well.

src/swarm/node/model/NeoNode.d
nemanja-boric-sociomantic: Correct.

PR #392, opened 2019-04-11
 Add new connection error notifier and error field to union

matthias-wende-sociomantic:
The user provided connection notifier was called on startup only in case
of an successfully established connection or if an error happened while
connecting. This patch adds a new (internal) client connection error
notifier that is called from the ISelectClient `error` method, in that
case the client Connection class. That happens when the ISelectClient is
unregistered from epoll due to a exception thrown in the RequestOnConn
fiber.

The exception is passed to the user by utilizing a new error field in
the ConnNotificationUnion (defined in the ConnectionSet).

### DIFF
--- a/src/swarm/neo/client/Connection.d
+++ b/src/swarm/neo/client/Connection.d
@@ -166,9 +166,18 @@ public final class Connection: ConnectionBase
 
     ***************************************************************************/
 
-    public alias void delegate ( Connection connection, Exception e = null ) StartupNotifier;
+    public alias void delegate ( Connection connection, Exception e = null ) Notifier;
 
-    private StartupNotifier startup_notifier = null;
+    private Notifier startup_notifier = null;
+
+    /***************************************************************************
+
+        Callback for notification when the an error happened
+        (`e` then reflects the error).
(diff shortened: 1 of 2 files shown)


gavin-norman-sociomantic:
Blocked on testing to confirm that the theory works in practice.

matthias-wende-sociomantic:
Changed base branch for testing. The actual patch should be merged into v6.x.x (i.e. the next major) though.

liza-churkyn-sociomantic:
Tested with my POC `UpdateRequest` app, specifically calling a blocking request from within the notifier of another blocking request. Doing that is a bug (as it leads to a wrong cast in the Task in the 2nd blocking request) and causes an error, and the thought behind this PR was to notify the client that an error has occurred by calling the Connection notifier with an error. But, the notifier is not called, and we don't reach  this line: https://github.com/sociomantic-tsunami/swarm/pull/392/files#diff-3fab881125f1bae0cbb5e9bd40b76b05R308.

Could it be that its because this is not an exception, but an error, and is not caught by the base code? Just speculating here...

PR #393, opened 2019-05-03
Remove deprecations

tiyash-basu-sociomantic:


src/swarm/neo/authentication/HmacAuthCode.d
```diff
@@ -109,10 +109,7 @@ struct HmacAuthCode
     import cstdio = core.stdc.stdio: fprintf, stderr;
     import core.stdc.stdlib: exit, EXIT_FAILURE;
 
-    version (D_Version2)
-        mixin("shared static this ( ) { staticCtor(); }");
-    else
-        static this ( ) { staticCtor(); }
+    shared static this ( ) { staticCtor(); }
```
gavin-norman-sociomantic: This could be refactored to include the code that's inside `staticCtor`. I don't think having that separate function is needed any more.

src/swarm/neo/node/Connection.d
```diff
@@ -23,7 +23,7 @@ class Connection: ConnectionBase
     import swarm.neo.node.RequestSet;
     import swarm.neo.protocol.connect.NodeConnect;
     import swarm.neo.authentication.HmacDef: Key;
-    import swarm.neo.connection.YieldedRequestOnConns;
+    public import swarm.neo.connection.YieldedRequestOnConns;
```
gavin-norman-sociomantic: Did you copy this decision (to use `public import`) from somewhere else (e.g. ocean)? It"s something we've traditionally avoided, but I may have missed that it's now being used in certain circumstances.

src/swarm/neo/client/ConnectionSet.d
```diff
@@ -917,12 +918,8 @@ version (UnitTest)
 unittest
 {
     static void setAddress ( ref AddrPort address, ubyte[] address_bytes ... )
-    in
```
gavin-norman-sociomantic, line 920: Are these changes related to the D2 transition?

src/swarm/client/registry/DisablableNodeSet.d
```diff
@@ -83,16 +83,14 @@ public class DisablableNodeSet : NodeSet
     ***************************************************************************/
 
     public void disable ( NodeItem node )
-    out
```
gavin-norman-sociomantic: Are these changes related to the D2 transition?

src/swarm/neo/node/IRequest.d
```diff
@@ -0,0 +1,44 @@
...
+    reason for this is documented in `ConnectionHandler.emplace`.
+
+    copyright: Copyright (c) 2018 dunnhumby Germany GmbH. All rights reserved
+
+    License:
+        Boost Software License Version 1.0. See LICENSE.txt for details.
+
+*******************************************************************************/
+
+module swarm.neo.node.IRequest;
```
gavin-norman-sociomantic, line 16: This rename is a breaking change and should have a release note (for the sanity of library users).

gavin-norman-sociomantic (commented):
A few comments, but looks good!

src/swarm/neo/node/Connection.d
tiyash-basu-sociomantic: This was not copied from any other project. I did not know that we are avoiding this.

Would you prefer to not use `public imports` here?

src/swarm/client/registry/DisablableNodeSet.d
tiyash-basu-sociomantic: This was mentioned in the D2-migration docs [here](https://github.com/sociomantic/backend/blob/master/doc/d2-migration/stage3.md#invariants).

src/swarm/neo/client/ConnectionSet.d
tiyash-basu-sociomantic: This was mentioned in the D2-migration docs [here](https://github.com/sociomantic/backend/blob/master/doc/d2-migration/stage3.md#invariants).

src/swarm/neo/authentication/HmacAuthCode.d
tiyash-basu-sociomantic: I added a release note for this.

src/swarm/neo/node/IRequest.d
tiyash-basu-sociomantic: I added a release note for this.

src/swarm/neo/authentication/HmacAuthCode.d
tiyash-basu-sociomantic: Done!

src/swarm/client/registry/DisablableNodeSet.d
gavin-norman-sociomantic: The more detailed doc about assert/enforce/verify says something different: https://github.com/sociomantic-tsunami/sociomantic/blob/master/Code/D/assert-vs-enforce.md#adapting-older-code
>All checks in in contracts should be converted to verify statements in the method/function body.
>out contracts and invariants can remain as they are, using asserts. They will not be present in release builds.

The weird thing is that we already did a sweep of removing all `in` contracts from swarm, so I'm confused as to how you've discovered some!

src/swarm/neo/node/Connection.d
gavin-norman-sociomantic: As far as I remember, in other library locations where we came across this, we added the imports in the locations where they're actually used.

src/swarm/client/registry/DisablableNodeSet.d
tiyash-basu-sociomantic: Interesting.. I have found some commits that remove `in` contracts, but the ones changed in this PR were not touched back then. Maybe they were missed for some reason?

Also, the `out` contracts have been put back now.

tiyash-basu-sociomantic:
Marked #390 to be fixed in the commit titled `Remove D1 tests from CI`.

src/swarm/neo/protocol/socket/MessageReceiver.d
```diff
@@ -383,17 +383,15 @@ private class MessageReceiverBase
...
-        assert(this); // invariant
-    }
     out
     {
         assert(bytes_requested <= this.buffer_content_end);
         assert(this); // invariant
     }
     body
     {
+        verify(this !is null);
```
gavin-norman-sociomantic: This is wrong. Doing `assert(this);` causes the invariant to be called. (Hence the comment.)

src/swarm/neo/protocol/socket/MessageReceiver.d
```diff
@@ -515,16 +513,14 @@ class MessageReceiver: MessageReceiverBase
...
-    {
-        assert(this);
-    }
     out
     {
         assert(this);
     }
     body
     {
+        verify(this !is null);
```
gavin-norman-sociomantic: Ditto.

src/swarm/client/registry/DisablableNodeSet.d
gavin-norman-sociomantic: It _could_ be that "in contracts (containing asserts) should only be used on private methods/ functions which are called via a public API including verify/enforce checks. in contracts on non-private methods/functions should be avoided." applies to them? (https://github.com/sociomantic-tsunami/sociomantic/blob/master/Code/D/assert-vs-enforce.md#general-error-checking-principles)

src/swarm/neo/node/Connection.d
tiyash-basu-sociomantic: I have reverted back to `import` from `public import` in most cases, and imported modules explicitly, wherever needed.

In some places in `swarm`, I noticed that `public import`s are already being used, e.g., in `swarm.util.Verify`. I also found a few instances in `ocean`. So, in the [commit in this pr](https://github.com/sociomantic-tsunami/swarm/pull/393/commits/23d6a2b93abe9ac1907f4d17b9a53bbdafb7ee05), `public import`s are still introduced in some places in tests, where adding them keeps the design simplified.

src/swarm/neo/protocol/socket/MessageReceiver.d
tiyash-basu-sociomantic: Thanks for pointing out. Replaced the invariant call with the invariant body here.

src/swarm/neo/protocol/socket/MessageReceiver.d
tiyash-basu-sociomantic: Replaced the invariant call with the invariant body here.

matthias-wende-sociomantic:
Does all of this have to happen in one PR. Otherwise I'd suggest to have one PR for pure D2 migration and a succeeding cleanup PR.

gavin-norman-sociomantic:
That's what I was wondering. We could have a PR for the D2 stuff, then a PR for the deprecations, say.

src/swarm/neo/client/RequestSet.d
```diff
@@ -606,7 +606,13 @@ public final class RequestSet: IRequestSet
         private void reset ( )
         out
         {
-            assert(this); // invariant
+            if (!this.id)
```
gavin-norman-sociomantic: Wait, what's the reason for this?

src/swarm/client/request/params/IRequestParams.d
```diff
@@ -538,7 +538,6 @@ public abstract class IRequestParams
         ***********************************************************************/
 
         static typeof(&this) serialized ( void[] data )
-        in
```
gavin-norman-sociomantic, line 541: The commit message for this commit (" Move `in` contracts within function bodies ") is a bit inaccurate. Maybe this?
"
Move `in` contracts within function bodies
All `in` contracts were converted into `verify` calls in method bodies in a previous PR. It seems some were missed. Converting these now. (The reason for converting `in` contracts to `verify` calls is to make important checks also occur in release builds.)
"

src/swarm/neo/client/RequestSet.d
tiyash-basu-sociomantic: Thanks for pointing out. It seems like I missed it. I'll do another sweep.

tiyash-basu-sociomantic: Ah, I see. I copied the contents of the invariant into the out block, so that checks specified here are not removed.

tiyash-basu-sociomantic:
As discussed, this PR has been split into two. This PR takes care of some optimizations post-D2 migration, and removing deprecations.

ATM, it's blocked on https://github.com/sociomantic-tsunami/swarm/pull/395 to drop D1 support.

tiyash-basu-sociomantic:
Rebased on the `v6.x.x` branch.

src/swarm/neo/protocol/socket/MessageReceiver.d
gavin-norman-sociomantic: Would be nice to not make this intermediary wrong change here.

src/swarm/neo/protocol/socket/MessageReceiver.d
gavin-norman-sociomantic: Would be nice to not make this intermediary wrong change here.

src/swarm/neo/client/RequestSet.d
gavin-norman-sociomantic: Why though? (This isn't what the assert/enforce/verify doc recommends doing. Unless I'm going mad.)

gavin-norman-sociomantic (commented):
All commits which I've not commented on look good.

tiyash-basu-sociomantic:
As discussed, this commit refactoring `out` contracts has been removed, so that invariants are still executed in non-release builds.

gavin-norman-sociomantic (approved):

Issue #303, opened 2018-03-08
`RequestOnConn.receive` should pass received data as `Const(void)[] data`


nemanja-boric-sociomantic:
Currently, `RequestOnConnBase.receive` passes the received data to the receive callback delegate as `in void[] data` (https://github.com/sociomantic-tsunami/swarm/blob/f337984496e0f08d08f4be5997b4272d9c0e0dab/src/swarm/neo/connection/RequestOnConnBase.d#L1029). This prevents the user changing this slice, so the user needs to do `Const!(void)[] payload_slice = payload` in order to be able to pass the data trough the `MessageParser.get*` methods, for example.

There's no need that the entire slice stay constant, it's has the scope of the delegate and it can safely be non-const. However, the data it points to should stay const, so it would be better if the delegate had the following signature:

```
void receive (Const(void)[] data)
```

gavin-norman-sociomantic:
Fixed in #397.

PR #397, opened 2019-05-09
Pass received data as `const(void)[]` from `EventDispatcher.receive`

tiyash-basu-sociomantic:
The delegate parameter of the `EventDispatcher.receive()` method now takes an array of const values, instead of a const array.

### DIFF
--- a/integrationtest/neo/client/request/internal/Get.d
+++ b/integrationtest/neo/client/request/internal/Get.d
@@ -150,9 +150,8 @@ public struct VersionedGet ( ubyte Version )
                             context.shared_working.result =
                                 SharedWorking.Result.Received;
 
-                            scope recv_cb = ( in void[] const_payload )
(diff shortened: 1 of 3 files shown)


gavin-norman-sociomantic (commented):
One tiny comment: could you copy some of the rationale for this change from #303 into the commit message? It's just always nice for the commit message to describe the reason for a change.

Re: testing, we'll test this thoroughly when we test *proto with the major release candidate of swarm.

tiyash-basu-sociomantic:
> One tiny comment: could you copy some of the rationale for this change from #303 into the commit message? It's just always nice for the commit message to describe the reason for a change.
> 
> Re: testing, we'll test this thoroughly when we test *proto with the major release candidate of swarm.

Thanks for pointing out. I've added the reasoning to the commit message now.

gavin-norman-sociomantic (approved):

PR #398, opened 2019-05-09
Return node's status from `client.RequestOnConn.useNode`

### LINKED ISSUE #372: Differentiate failure reasons of `swarm.client.RequestOnConn.useNode`
`useNode` method fails for two reason: if the node is not in the connection set (if it was never connected/contacted for example) and if the node is not currently connected. Both these failures represent as `return false` error code. This makes impossible for clients to enforce different policies (for example, `dhtproto` could use `no_node` vs `node_disconnected` error to differentiate between these two error codes).

See also: https://github.com/sociomantic-tsunami/swarm/pull/371

tiyash-basu-sociomantic:
This helps the caller check between success and failure of the connection,
and also distinguish between two different failure reasons,
a. between a node being absent, or
b. there being a connection failure to the said node.

Fixes #372

### DIFF
--- a/src/swarm/neo/client/RequestOnConn.d
+++ b/src/swarm/neo/client/RequestOnConn.d
@@ -331,6 +331,24 @@ public class RequestOnConn: RequestOnConnBase, IRequestOnConn
 
     /***************************************************************************
 
+        Enumerated states of nodes to indicate their connection statuses.
+
+    ***************************************************************************/
+
+    public enum NodeState
+    {
+        /// The node is connected
+        Connected,
+
+        /// The node does not exist
+        Missing,
+
+        /// The node is disconnected
+        Disconnected
+    }
+
+    /***************************************************************************
+
         Exception class thrown to abort a request-on-conn. Exceptions of this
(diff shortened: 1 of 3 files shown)


matthias-wende-sociomantic (approved):

src/swarm/neo/client/RequestOnConn.d
```diff
@@ -329,6 +329,24 @@ public class RequestOnConn: RequestOnConnBase, IRequestOnConn
...
+    public enum NodeState
+    {
+        /// The node is connected
+        Connected,
+
+        /// The node does not exist
+        Missing,
+
+        /// The node is disconnected
+        Disconnected
```
gavin-norman-sociomantic, line 347: This did make me wonder what the request handler is recommended to do in the case where the node is known but disconnected. I know that all-nodes requests typically automatically wait for the connection to be reestablished, but this isn't the normal behaviour in single-node requests. I think the norm there is to simply return an error to the user and expect the user to decide what to do (i.e. retry or abandon the request).

No change required here, just noting my thoughts, for fun :)

PR #399, opened 2019-05-16
Fixes to release candidate v6.0.0-rc.1

tiyash-basu-sociomantic:
The issues addressed in the commits here were identified while testing swarm `v6.0.0-rc.1` with dhtproto.


src/swarm/client/RequestSetup.d
```diff
@@ -718,6 +718,24 @@ public template Suspendable ( )
...
+
+        Copies the value of the suspend unregister member into the provided
+        request params class instance.
+
+        Params:
+            params = IRequestParams instance to write into
+
+    ***************************************************************************/
+
+    private void setup_suspend_unregister ( IRequestParams params )
```
gavin-norman-sociomantic, line 732: Was this removed by accident or missing?

tiyash-basu-sociomantic: Turns out this was accidentally removed by me in https://github.com/sociomantic-tsunami/swarm/commit/6e26ffa.

tiyash-basu-sociomantic: This method had a comment `// deprecated: remove in swarm v6`.
This means that the commit adding this in this PR should be removed, and dhtproto should be fixed instead?

gavin-norman-sociomantic: Looks to me like it meant "remove the static if". (That's without looking into things more deeply.)

gavin-norman-sociomantic: If it'd been me writing that deprecation comment (which it most likely was), I would have written it in the method doc if I'd meant to remove the whole method ;)

tiyash-basu-sociomantic: The tests in dhtproto pass now, with the static if removed.

tiyash-basu-sociomantic: For now, I have updated the commit description to better indicate the change.

PR #401, opened 2019-05-16
Remove deprecated static ifs in RequestMap.addHandler

gavin-norman-sociomantic:


tiyash-basu-sociomantic:
LGTM. All tests pass with dhtproto.

matthias-wende-sociomantic:
Same idea as Tiyash had (https://github.com/sociomantic-tsunami/dlsproto/pull/110#discussion_r285024023). We might want to add (final) getter/setter functions to the `IRequest` interface to enforce derived classes to add the `timing` and `scheduled_for_removal` members.

gavin-norman-sociomantic:
The reason for not being able to add these fields to the interface is that they're `static`. They're used at run-time in the `addHandler` template to check for certain properties of the request type.

PR #403, opened 2019-09-06
Remove use of deprecated FieldName template

matthias-wende-sociomantic:
FieldName is long deprecated in ocean and was
removed in the latest major version.


don-clugston-sociomantic:
LGTM

PR #405, opened 2019-10-20
Add test support for Ubuntu bionic

daniel-zullo:
Update beaver build image to support Ubuntu xenial and bionic distributions and add support for Ubuntu bionic.


daniel-zullo:
This is now blocked due to the issue reported in https://issues.dlang.org/show_bug.cgi?id=20441

Issue #406, opened 2019-10-28
Stop using templates Const, Immut and Inout 


don-clugston-sociomantic:
The templates Const, Immut and Inout were used for transitioning from D1 to D2. Now that swarm has been converted, they should be replaced by plain D2 keywords.

jens-mueller-sociomantic:
Closed by https://github.com/sociomantic-tsunami/swarm/pull/407.

PR #407, opened 2019-10-29
Replace all usages of `Const!` and `Immut!`

jens-mueller-sociomantic:
With this change all usage of the templates `Const` and `Immut` is replaced by
their appropriate D2 builtins.
The code base is now free of `Const`, `Immut` and `Inout`. For `Inout` no
changes had to be made because it is not used `Inout`.

Fixes sociomantic-tsunami/swarm#406.


don-clugston-sociomantic (approved):
LGTM

jens-mueller-sociomantic:
Just to make sure: Am I targeting the right branch?

jens-mueller-sociomantic:
Shouldn't I target `v5.x.x`?

jens-mueller-sociomantic:
Just checked: I think v6.x.x is fine because this should be when D1 support was dropped.

jens-mueller-sociomantic:
How many approvals do I need to merge it? I remember with some libraries you need at least two.

alaksiej-stankievicx-dunnhumby (approved):

PR #410, opened 2019-12-19
Replace deprecated D1 operator overloads

don-clugston-sociomantic:
To reduce compilation warnings when compiling with dmd 2.088 and later.
These are the simple cases, where the operators are members of structs, so the fact that the operators are non-virtual is not a problem.


ben-palmer-sociomantic (approved):
LGTM

PR #412, opened 2020-01-27
Remove use of genOpCmp, genOpEquals mixins

don-clugston-sociomantic:
These mixins were only required during D2 transition, and can now be removed.


stefan-koch-sociomantic (approved):
Looks good, nice work.

PR #413, opened 2020-01-31
Replace all imports of ocean.transition

don-clugston-sociomantic:
The ocean.transition module was created to ease D2 conversion. It can now be replaced with more appropriate imports.

The first commit replaces `mixin TypeofThis` with `typeof(this)`. The second commit deals with the one module which required the transition import to be replaced with two imports.
The third commit is just a search-and-replace of `import ocean.transition`.


daniel-zullo (approved):
LGTM

PR #414, opened 2020-03-12
Add annotations for fns which `return &this`

don-clugston-sociomantic:
From dmd2.092 onwards, functions which use the `return &this` method chaining idiom, must be marked with the `return` annotation. This makes it impossible for @safe code to have references to destructed stack objects (see DIP25).

Fixes deprecation messages which occur with dmd2.092 beta.


Geod24 (approved):
Thanks for keeping up with upstream!

PR #416, opened 2020-03-23
Add -dip25 to Build.mak, to check dmd 2.092

don-clugston-sociomantic:
DMD 2.092 has not actually been released yet, but the main breaking
change it has it that -dip25 is enabled. So -dip25 should be set for
all supported compilers.


don-clugston-sociomantic:
This is not complete, it needs my other ocean PR to be merged first

Geod24:
Since `-dip25` is the default in v2.093, we can just close this.

PR #420, opened 2020-03-27
Remove usage and mentions of `ocean.stdc`

Geod24:
All the functions and symbols used here have long been in druntime and can be used instead of the thin wrapper in ocean.

### DIFF
--- a/src/swarm/neo/client/RetryTimer.d
+++ b/src/swarm/neo/client/RetryTimer.d
@@ -14,6 +14,7 @@ module swarm.neo.client.RetryTimer;
 
 import core.stdc.errno;
 import stdio = core.stdc.stdio;
+import core.stdc.string;
 import core.sys.posix.sys.time;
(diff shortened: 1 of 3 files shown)


don-clugston-sociomantic:
./src/swarm/neo/client/RetryTimer.d(17): Error: module string is in file 'core/sys/posix/string.d' which cannot be read

Should that be `core.stdc.string` ?

Geod24:
No, the module is too recent (2.089): https://github.com/dlang/druntime/commit/a2a88cf05cccbf70e7a1df167d28a45b827da5d4#diff-da1264dc2f78bd6c9e7fe0b1508f932f

Geod24:
Well actually it seems to be only using `strerror` so I'll try `core.stdc.string`

Geod24:
Ping @don-clugston-sociomantic

don-clugston-sociomantic:
LGTM

PR #422, opened 2020-08-13
Remove v2.071.2.s* from the test matrix

Geod24:
This compiler is long gone and there is no interest in supporting it in the future.
This is the lowest hanging fruit in the matrix and will allow upstream to move
forward immediately.


Geod24:
Self merging following approval on Ocean: https://github.com/sociomantic-tsunami/ocean/pull/829

Issue #417, opened 2020-03-23
Neo requests escape a reference to a stack variable.


don-clugston-sociomantic:
When compiled with -dip25, dmd give an alarming warning:

./submodules/swarm/src/swarm/neo/connection/RequestOnConnBase.d(416): Deprecation: copying (cast(const(void*))&elem)[0..4] into allocated memory escapes a reference to parameter variable elem

The code is:
```
            public void add ( T ) ( ref T elem )
            {
                static assert(!hasIndirections!(T));
                this.outer.outer.send_payload ~= (cast(const(void*))&elem)[0..T.sizeof];
            }
```
And actually I'm horrified that such a nasty bit of code has no comments.
Now look at what this is doing. Suppose T is an int. This creates a `void[]` slice of the raw bytes of elem, and appends them to `send_payload`. That's a hacky way of serializing but OK.

But here's the problem. `send_payload` is not an array of bytes! It is actually:
``` 
         private const(void)[][] send_payload;
```
So though it looks as though it appends the bytes, that's not what it's doing. It's appending the slices. Ie it is saving **pointers to stack variables**.

Now it looks as though in practice, the send_payload array is used immediately after being set, and then the array is cleared.
However, in most cases, the parent class, and hence the array itself, has a lifetime which is longer than the stack variables which it points to. It's possible that all extant uses are correct in the sense that they avoid memory corruption. But it all seems rather fragile.

 I am not sure how to deal with this.

don-clugston-sociomantic:
There is a second escaping reference in `appendSlices` in src/swarm/neo/util/Util.d

It is basically the same code, though it has code comments which indicates that's it's hacky. However, it appears that this function is never actually used anywhere in swarm or dmq/dls/dhtproto.

Geod24:
> Ie it is saving pointers to stack variables.

Little correction: The argument is `ref`, so it's saving a pointer to the caller's frame. It is indeed very fragile.
There is `addCopy` with takes a value and will perform a "deep" copy for integer types.

I think the issue here was that they wanted an array of disjoint types with no overhead.

Geod24:
The deprecation is fixed. The usage, not so much, but that's for another issue.

PR #428, opened 2020-08-25
Silence in `RequestOnConnBase.Payload.add` & co

Geod24:
See each commit / the release notes for explanation. Unfortunately I also found a regression.
This should be the last stone to get v6.1.0 (and v7.0.0) released.

### DIFF
--- a/src/swarm/neo/connection/RequestOnConnBase.d
+++ b/src/swarm/neo/connection/RequestOnConnBase.d
@@ -412,8 +412,26 @@ abstract class RequestOnConnBase
 
             public void add ( T ) ( ref T elem )
             {
+                this.addPointer(&elem);
+            }
+
+            /*******************************************************************
+
+                Adds a single element to the payload to be sent.
+
+                Note that this takes a pointer to the element, and the element's
+                lifetime needs to match or exceed that of this payload to avoid
+                memory corruption.
+
+                Params:
+                    ptr = pointer to the element to add
+
+            *******************************************************************/
+
+            public void addPointer ( T ) ( const T* ptr )
+            {
                 static assert(!hasIndirections!(T));
-                this.outer.outer.send_payload ~= (cast(const(void*))&elem)[0..T.sizeof];
+                this.outer.outer.send_payload ~= (cast(const(void*))ptr)[0..T.sizeof];
             }
 
             /*******************************************************************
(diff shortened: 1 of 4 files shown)


ibuclaw (approved):
I agree, it looks very dubious.  Maybe wait for a second review from @joseph-wakeling-frequenz or @ben-palmer-sociomantic though.

Geod24:
Fixes https://github.com/sociomantic-tsunami/swarm/issues/417

src/swarm/neo/connection/RequestOnConnBase.d
```diff
@@ -411,9 +411,27 @@ abstract class RequestOnConnBase
...
+            /*******************************************************************
+
+                Adds a single element to the payload to be sent.
+
+                Note that this takes a pointer to the element, and the element's
+                lifetime needs to match or exceed that of this payload to avoid
+                memory corruption.
+
+                Params:
+                    ptr = pointer to the element to add
```
daniel-zullo, line 427: it should be `elem` instead. Otherwise the input parameter in the code needs to be updated

src/swarm/neo/connection/RequestOnConnBase.d
```diff
@@ -411,9 +411,27 @@ abstract class RequestOnConnBase
...
+
+            *******************************************************************/
+
+            public void addPointer ( T ) ( const T* elem )
```
daniel-zullo: Ideally it should overload the function `add`  with different input parameter as both functions share the same functionality.

daniel-zullo (commented):
A couple of suggestions and LGTM

src/swarm/neo/connection/RequestOnConnBase.d
Geod24: I think we should deprecate / remove add in the future. It exposes too much a dangerous interface.
This way we'll just have `addCopy` / `addPointer` / `addArray` (which calls `addCopy`).
Perhaps a more descriptive name can be used, such as `addScopedData` ?

Note: Turning a function into an overload set can break code, too.

src/swarm/neo/connection/RequestOnConnBase.d
Geod24: Good catch!

src/swarm/neo/connection/RequestOnConnBase.d
daniel-zullo: Then it's fine to keep it as it is and no need to change the name.

daniel-zullo (approved):

PR #443, opened 2025-09-17
[protocol/FiberSelectWriter.d] fix null deref in not run unittest cod…

thewilsonator:
…e example

PR #444, opened 2025-09-22
fix build error due to missing ctor arguments in #443

rainers:
to unblock the dmd test suite, see also comments in #443.
@Geod24 @thewilsonator

### DIFF
--- a/src/swarm/protocol/FiberSelectWriter.d
+++ b/src/swarm/protocol/FiberSelectWriter.d
@@ -105,7 +105,7 @@ public class FiberSelectWriter : Ocean.BufferedFiberSelectWriter
     {
         void instantiate ()
         {
-            auto writer = new FiberSelectWriter;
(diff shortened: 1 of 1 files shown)


kubo39:
I'm not sure, but should line 147 be fixed too?

rainers:
> I'm not sure, but should line 147 be fixed too?

I was also wondering why it hasn't been changed in #443, but I didn't want to expand the scope of this PR.

src/swarm/protocol/FiberSelectWriter.d
```diff
@@ -105,7 +105,7 @@ public class FiberSelectWriter : Ocean.BufferedFiberSelectWriter
     {
         void instantiate ()
         {
-            auto writer = new FiberSelectWriter;
+            auto writer = new FiberSelectWriter( null );
```
Geod24:
```suggestion
            auto writer = new FiberSelectWriter(null);
```

